PR #3, opened 2015-02-03
librina: fixing several concurrency.h issues

msune:


msune:
So this fixes are coming batched in https://github.com/bisdn/rina-sdk/tree/wip , also to support dana-i2cat/irati#450

So we could hold this pull req and wait for the whole batch there

PR #484, opened 2015-03-11
Pristine1.1/rwlock

msune:
Adding Scoped rwlocks (RAI)


msune:
@salvestrini @ all if you can review this patch and integrate asap, because this is blocking me here:

https://github.com/bisdn/irati/tree/pristine1.1/ipcm_refactor

PR #486, opened 2015-03-12
vmpi: fix some include directives

vmaffione:
Fix some include directives to allow compilation if VMPI_TX_MUTEX is defined.


salvestrini:
master-1.0.0 considered frozen.

PR #498, opened 2015-03-23
Simplyfing logging facility. Adding likely.h header (branch prediction)

msune:
* Substituted rwlocks for a single mutex on logging stream setting
* Removing unnecessary (and expensive) double locking per trace call


msune:
superseded (requested in pristine-1.2)

Issue #499, opened 2015-03-23
kernel: Query RIB command returns an invalid NL message


msune:
Using pristine-1.1 `HEAD` (0163aa356ce5f3bf8beafad1253623618ef99016), issue a Query RIB command over a dummy shim, the returned NL message contains invalid version and code. 

There seems to be rubbish at the very beginning of the number (25->255, 22->222):

```
 822(1427121237)#netlink-manager (DBG): Received 88 bytes, parsing the message
822(1427121237)#netlink-manager (DBG): Source: 0, Netlink family: 2; Version: 255; Operation code: 222; Flags: 0; Sequence number: 1427121218
822(1427121237)#netlink-parsers (ERR): Generic Netlink message contains unrecognized command code: 222
822(1427121237)#netlink-manager (ERR): Error parsing Netlink message
822(1427121237)#netlink-manager (DBG): Sent message of 68 bytes. Family: 25; Operation code: 22_QUERY_RIB_REQUEST; Source port: 1; Destination port: 0; Sequence Number: 1427121218
Is request message? 1; Is response message? 0; Is notification message? 0
Source IPC Process: 0; Destination IPC Process: 1
822(1427121237)#ipcm (INFO)[query_rib]: Requested query RIB of IPC process shim1:1::

822(1427121237)#core (ERR): Error receiving netlink message. Error parsing Netlink message
822(1427121237)#core (DBG): Waiting for message 1
822(1427121246)#ipcm (ERR)[query_rib]: : Timed out
```

edugrasa:
@vmaffione this issue can be closed

PR #505, opened 2015-03-24
Pristine 1.2 #499 Added handling of query RIB operation to shim IPCPs in the kernel

edugrasa:
Fixing issue #499, right now the default behaviour is to return an error since the operation is not implemented yet (but this is enough since a proper NL response message is provided to the IPC Manager).


sandervrijders:
1) Please fix the identation 
2) Use LOG_MISSING instead of //TODO
3) Don't use C++ style comments
4) Don't combine declarations with initializations  
5) Please add error messages whenever something fails (line 1325 for example)

Please re-issue a pull request when these comments are addressed. Thanks!

PR #515, opened 2015-03-27
librina, rinad: add -std=c++98 compiler option

vmaffione:
Add compiler flags to force user-space stack to follow the c++98 standard.

### DIFF
--- a/librina/src/Makefile.am
+++ b/librina/src/Makefile.am
@@ -53,6 +53,7 @@ librina_la_CPPFLAGS =				\
 	$(LIBPROTOBUF_CFLAGS)			\
 	$(CPPFLAGS_EXTRA)
 librina_la_CXXFLAGS =				\
+	-std=c++98				\
 	$(CXXFLAGS_EXTRA)
 
 librina_stubbed_la_SOURCES  = $(allsources)
@@ -70,6 +71,7 @@ librina_stubbed_la_CPPFLAGS =			\
 	$(LIBPROTOBUF_CFLAGS)			\
 	$(CPPFLAGS_EXTRA)
 librina_stubbed_la_CXXFLAGS =			\
+	-std=c++98				\
 	$(CXXFLAGS_EXTRA)
 
 lib_LTLIBRARIES    = librina.la
--- a/rinad/src/common/Makefile.am
+++ b/rinad/src/common/Makefile.am
@@ -15,6 +15,9 @@ librinad_la_CPPFLAGS =				\
 	$(LIBPROTOBUF_CFLAGS)			\
 	-I$(srcdir)/tclap			\
 	-I$(srcdir)/..
+librinad_la_CXXFLAGS =				\
+	-std=c++98				\
+	$(CXXFLAGS_EXTRA)
 librinad_la_LIBADD   =				\
 	$(LIBRINA_LIBS)				\
 	$(LIBPROTOBUF_LIBS)			\
--- a/rinad/src/ipcm/Makefile.am
+++ b/rinad/src/ipcm/Makefile.am
@@ -53,6 +53,9 @@ ipcm_CPPFLAGS =					\
 	-I$(srcdir)/../common			\
 	-I$(srcdir)/jsoncpp			\
 	-DPLUGINSDIR=\"$(pkglibdir)/ipcm\"
+ipcm_CXXFLAGS =					\
+	-std=c++98				\
+	$(CXXFLAGS_EXTRA)
 ipcm_LDADD    =					\
 	$(builddir)/../common/librinad.la	\
 	$(builddir)/libjsoncpp.la		\
@@ -78,6 +81,9 @@ test_empty_CPPFLAGS =				\
 	$(LIBRINA_CFLAGS)			\
 	-I$(srcdir)/..				\
 	-I$(srcdir)/../common
+test_empty_CXXFLAGS =				\
+	-std=c++98				\
+	$(CXXFLAGS_EXTRA)
 test_empty_LDADD    = $(builddir)/../common/librinad.la $(LIBRINA_LIBS)
 
 check_PROGRAMS =				\
--- a/rinad/src/ipcp/Makefile.am
+++ b/rinad/src/ipcp/Makefile.am
@@ -25,6 +25,9 @@ ipcp_CFLAGS   =
 ipcp_CPPFLAGS =					\
 	$(COMMONCPPFLAGS)			\
 	-DPLUGINSDIR=\"$(pkglibdir)/ipcp\"
+ipcp_CXXFLAGS =					\
+	-std=c++98				\
+	$(CXXFLAGS_EXTRA)
 ipcp_LDADD    = $(COMMONLIBS) -ldl
 ipcp_SOURCES  =						\
 	main.cc						\
@@ -48,6 +51,9 @@ plugins_LTLIBRARIES =
 default_la_CFLAGS =
 default_la_CPPFLAGS =				\
 	$(COMMONCPPFLAGS)
+default_la_CXXFLAGS =				\
+	-std=c++98				\
+	$(CXXFLAGS_EXTRA)
 default_la_LIBADD = $(COMMONLIBS) -ldl
 default_la_LDFLAGS = -module
 default_la_SOURCES =				\
@@ -71,6 +77,9 @@ test_encoders_SOURCES  =			\
 	enrollment-task.cc   enrollment-task.h	\
     link-state-routing-ps.cc   link-state-routing-ps.h
 test_encoders_CPPFLAGS = $(testsCPPFLAGS)
+test_encoders_CXXFLAGS =			\
+	-std=c++98				\
+	$(CXXFLAGS_EXTRA)
 test_encoders_LDADD    = $(testsLIBS)
 
 test_pduftg_SOURCES  =				\
@@ -79,6 +88,9 @@ test_pduftg_SOURCES  =				\
 	events.cc          events.h		\
 	link-state-routing-ps.cc link-state-routing-ps.h
 test_pduftg_CPPFLAGS = $(testsCPPFLAGS)
+test_pduftg_CXXFLAGS =				\
+	-std=c++98				\
+	$(CXXFLAGS_EXTRA)
 test_pduftg_LDADD    = $(testsLIBS)
 
 check_PROGRAMS =				\
--- a/rinad/src/ipcp/plugins/Makefile.am
+++ b/rinad/src/ipcp/plugins/Makefile.am
@@ -19,6 +19,9 @@ plugins_LTLIBRARIES =
 sm_passwd_la_CFLAGS   =
 sm_passwd_la_CPPFLAGS =				\
 	$(COMMONCPPFLAGS)
+sm_passwd_la_CXXFLAGS =				\
+	-std=c++98				\
+	$(CXXFLAGS_EXTRA)
 sm_passwd_la_LIBADD   =
 sm_passwd_la_LDFLAGS  = -module
 sm_passwd_la_SOURCES  =				\


msune:
I think it is better to add the `CFLAGS`/`CXXFLAGS` in the `configure.ac`'s of the different packages.

I would also imply --enable-debug to toggle `Wall` and `Werror`

vmaffione:
@msune, you mean putting the -std=c++98 flag just at configure.ac level - and not putting it in the single Makefile.am files?

vmaffione:
@msune, you mean putting the -std=c++98 flag just at configure.ac level - and not putting it in the single Makefile.am files?

msune:
Right

salvestrini:
CFLAGS and CXXFLAGS should not be overridden in configure.ac. We must use per-binary/library flags.

[ c++98, Wall and Werror are GCC specific (and thus these flags support should be checked also) ]

We should have a package-wide variables set (COMMON_[CPPFLAGS, CFLAGS and CXXFLAGS]) to add all such flags in per-binary/library fashion

msune:
>>FLAGS and CXXFLAGS should not be overridden in configure.ac. We must use per-binary/library flags.

c++98 is a very general option that I strongly doubt you want to define per library or per binary (e.g. define differently within librina). For the rest, e.g. Wall and Werror, it is fine to do it conditionally based to the compiler, but I doubt you also want to define them on a per binary/library.

(
Btw clang and icc are 100% compatible with GCC options, and specially with these generic ones; what compiler do you have in mind other than those to conditionally add them?
)

It is ok to not set it for releases, i.e. so that users don't stumble with stupid compilation errors like unused variable, sincedifferent GCC versions have different tolerances to Wunused for instance. But other than that, I don't see the point.

For other specifics, such as app/lib specific macros etc.., I agree with the COMMON_ and specific flags approach

salvestrini:
Flags are (in general) per compiler specific, depending on gcc/icc or whatever is just bad approach. Moreover, checking them with a macro is just straightforward and adds no complexity at all (few mins, just to drop in there an autom4te macro and call it. Thus, we can simply avoid that at no cost.

These flags have to be removed upon release for sure (i.e. don't stumble upon compilation errors due to different versions of toolings available on build machines)

msune:
I don't get the point. `configure` is there to help the user. 99.9% of the users should use some default flags, and a huge percentage of the current developers _should_ too. Not having it by default in the configure.ac is prone to errors, and certainly not an option for end users who only want to compile it. I don't like the idea that a developer is not setting the `-std=c++98`, for instance, in the `CFLAGS/CXXFLAGS` by mistake, commits, pushes and after someone else with that flags compiles it gives the alert, and developer has to come back, fix it and resend the pull req. This is a waste of time.

However, by no means I am saying the `CFLAGS/CXXFLAGS...` should be hardcoded or ignore user's one. Never did. What I am saying, like with the `--enable-debug` discussion, is that certain  "defaults" are **very** handy. This is why I am generally appending the user `CFLAGS/CXXFLAGS...`, to the end of configure's flags, so that user flags have preference. We could even have a switch, for experienced users in `configure`, something like `--without-default-flags` which completely disables them.

Having said that, and before we over-engineer the thing, I will reformulate the question; are we planning to support any _other_ compiler that is not gcc, or a drop-in replacement like clang or icc (that are 100% compatible, flags-wise)?

`CFLAGS/CXXFLAGS...` are in itself GCC specific environment variables. All the drop-in replacements listed have adopted it, sure, the same way they have adhered to support the same flags hence being drop-in replacements. I struggle to understand your reluctance to tweak `configure.ac`

salvestrini:
2015-03-30 14:10 GMT+02:00 Marc Suñé <notifications@github.com>:

> I don't get the point. configure is there to help the user. 99.9% of the
> users should use some default flags, and a huge percentage of the current
> developers *should* too. Not having it by default in the configure.ac is
> prone to errors, and certainly not an option for end users who only want to
> compile it. I don't like the idea that a developer is not setting the
> -std=c++98, for instance, in the CFLAGS/CXXFLAGS by mistake, commits,
> pushes and after someone else with that flags compiles it gives the alert,
> and developer has to come back, fix it and resend the pull req. This is a
> waste of time.
>
This is broken, we know.

> However, by no means I am saying the CFLAGS/CXXFLAGS... should be
> hardcoded or ignore user's one. Never did.
>
For sure, since they *MUST* no be hardcoded.

> What I am saying, like with the --enable-debug discussion, is that
> certain "defaults" are *very* handy. This is why I am generally appending
> the user CFLAGS/CXXFLAGS..., to the end of configure's flags, so that
> user flags have preference. We could even have a switch, for experienced
> users in configure, something like --without-default-flags which
> completely disables them.
>
A discussion I'm not aware of (or I slipped over unwillingly). However, an
--enable-developer-mode was planned (without "discussions") [and implicitly
stated in a previous response of mine].

> Having said that, and before we over-engineer the thing, I will
> reformulate the question; are we planning to support any *other* compiler
> that is not gcc, or a drop-in replacement like clang or icc (that are 100%
> compatible, flags-wise)?
>
One autom4te if is not overengineering and would suit all the environments
(is the flag supported [regardless of compiler and version] ? yeah -> use
it [if requested by the "mode"]). It's a 5 mins change.

> CFLAGS/CXXFLAGS... are in itself GCC specific environment variables.All
> the drop-in replacements listed have adopted it, sure, the same way they
> have adhered to support the same flags hence being drop-in replacements. I
> struggle to understand your reluctance to tweak configure.ac
>
I'm reluctant to crap up configure.ac, not to change it to suit the needs
(in a clean way) and IMO, the clean way is:

1) test if the tool supports the flag (e.g. using an autom4te macro)
2) check if the user wants to have it (e.g. using an
--enable-developer-mode/--disable-developer-mode flag from configure)
3) add it (to COMMON_*FLAG, used by <binary>_[CPP|C|CXX]FLAGS etc.)

takes a little and it is generic. No? Hardcoding flags would only cause
problems.

Ciao,
Francesco



> —
> <https://github.com/IRATI/stack/pull/515#issuecomment-87652591>.
>



-- 
Francesco Salvestrini
neXtworks s.r.l.
Via Livornese, 1027
56122 S. Piero a Grado (PI) - ITALY
Phone: +39 050 3871647 / Fax: +39 050 3871601

msune:
(You didn't answer whether we plan to support other compilers)

My original concern was not to pollute `Makefile.am` with specific flags, such as `-std=c++98`. I think you and me agree on this, isn't it?

I am fine with the approach of 1) 2) and 3). I think it is slightly over engineered for what we need, but I am OK with it. I agree is probably cleaner, than what I was proposing.

So, wrt to the PR, do we agree that this PR is NACK?

Next, we need `--enable-developer-mode` or whatever (I called this `--enable-debug`); do we open an issue for this? Who takes care of this?

salvestrini:
2015-03-30 16:08 GMT+02:00 Marc Suñé <notifications@github.com>:

> (You didn't answer whether we plan to support other compilers)
>
It's implicit: why not ?


> My original concern was not to pollute Makefile.am with specific flags,
> such as -std=c++98. I think you and me agree on this, isn't it?
>
configure.ac -> checks, AC_SUBST and then use the var in Makefile.am. No
pollution


> I am fine with the approach of 1) 2) and 3). I think it is slightly over
> engineered for what we need, but I am OK with it. I agree is probably
> cleaner, than what I was proposing.
>
> So, wrt to the PR, do we agree that this PR is NACK?
>

NACK this one

> Next, we need --enable-developer-mode or whatever (I called this
> --enable-debug); do we open an issue for this? Who takes care of this?
>
 Let's add an issue

Ciao,
Francesco

> —
> <https://github.com/IRATI/stack/pull/515#issuecomment-87692819>.
>



-- 
Francesco Salvestrini
neXtworks s.r.l.
Via Livornese, 1027
56122 S. Piero a Grado (PI) - ITALY
Phone: +39 050 3871647 / Fax: +39 050 3871601

msune:
>> (You didn't answer whether we plan to support other compilers)

> It's implicit: why not ?

Which one? Just curious...

>> My original concern was not to pollute Makefile.am with specific flags,
 such as -std=c++98. I think you and me agree on this, isn't it?

> configure.ac -> checks, AC_SUBST and then use the var in Makefile.am. No
pollution

Agree. This is not what is in this PR => NACK.

vmaffione:
If no more discussion is to happen, I will close this (not merging).

msune:
Close please

salvestrini:
Don't use fprintf
Don't use exit()

2015-04-08 10:45 GMT+02:00 Vincenzo Maffione <notifications@github.com>:

> Closed #515 <https://github.com/IRATI/stack/pull/515>.
>
> —
> <https://github.com/IRATI/stack/pull/515#event-276023882>.
>



-- 
Francesco Salvestrini
neXtworks s.r.l.
Via Livornese, 1027
56122 S. Piero a Grado (PI) - ITALY
Phone: +39 050 3871647 / Fax: +39 050 3871601

msune:
@salvestrini don't get it

Was this comment for https://github.com/IRATI/stack/pull/529?

And if it was, then why can we not use the std library

salvestrini:
I've surely messed up the reply. Was addressing #529

2015-04-08 11:50 GMT+02:00 Marc Suñé <notifications@github.com>:

> @salvestrini <https://github.com/salvestrini> don't get it
>
> Was this comment for #529 <https://github.com/IRATI/stack/pull/529>?
>
> And if it was, then why can we not use the std library
>
> —
> <https://github.com/IRATI/stack/pull/515#issuecomment-90865820>.
>



-- 
Francesco Salvestrini
neXtworks s.r.l.
Via Livornese, 1027
56122 S. Piero a Grado (PI) - ITALY
Phone: +39 050 3871647 / Fax: +39 050 3871601

msune:
(copy&pasting my answer by Skype):

@salvestrini : just to clarify. If this comment:

https://github.com/IRATI/stack/pull/515#issuecomment-90846330

Was for PR: https://github.com/IRATI/stack/pull/529, as I guess.

a) return and exit are equivalent in a main  routine.
b) fprintf is intentionally used, because I don't want the logger of librina to swallow this trace, no matter which default value of debug level is there.

salvestrini:
2015-04-08 12:15 GMT+02:00 Marc Suñé <notifications@github.com>:

> (copy&pasting my answer by Skype):
>
> @salvestrini <https://github.com/salvestrini> : just to clarify. If this
> comment:
>
> #515 (comment)
> <https://github.com/IRATI/stack/pull/515#issuecomment-90846330>
>
> Was for PR: #529 <https://github.com/IRATI/stack/pull/529>, as I guess.
>

It was

> a) return and exit are equivalent in a main routine.
>

Yes, but why mixin' them ?

> b) fprintf is intentionally used, because I don't want the logger of
> librina to swallow this trace, no matter which default value of debug level
> is there.
>
ERROR shouldn't be swallowed. Otherwise Use a CRITICAL level (that must not
swallow anything). Otherwise it would be ... just  messin' up

—
> <https://github.com/IRATI/stack/pull/515#issuecomment-90870846>.
>



-- 
Francesco Salvestrini
neXtworks s.r.l.
Via Livornese, 1027
56122 S. Piero a Grado (PI) - ITALY
Phone: +39 050 3871647 / Fax: +39 050 3871601

msune:
Actually, to properly clean resources `exit()` should be preferred instead of `return`, so the other returns should be the ones to be changed. That's not messing up, that's improving ;).

For the `fprintf()`, as you say, it _should_, but I don't want to depend on the logger implicit initialization. This check is the very first one that has to be made, even before attempting to parse arguments to deduce the logging level or anything, and we need this trace to pop up **always**.

The logger is statically and implicitly pre-initialized (which I don't like => static initialization errors are difficult to debug and act upon), but the actual level and stream are set during the body of the program. I don't want to depend on any implicit initialization of the logger, because root permissions is the very first thing we need to check, even before configuring the logger. 

Something like this is what I would expect:
```
//check root (and use fprintf(stderr, "") in purpose)
//Parse arguments
//Initialize logger
//here **use** the logger, not before with unknown(implicit) initialization
```

In any case, this commit is merged. This review is late.

salvestrini:
2015-04-08 12:49 GMT+02:00 Marc Suñé <notifications@github.com>:

> Actually, to properly clean resources exit() should be preferred instead
> of return, so the other returns should be the ones to be changed. That's
> not messing up, that's improving ;).
>
No, it is messing. a fallthrough of returns and one exit. Using returns,
destructors (locally scoped) will be called. Using exit(), they won't.
That's why there is (was) 1 single exit at most.

> For the fprintf(), as you say, it *should*, but I don't want to depend on
> the logger implicit initialization. This check is the very first one that
> has to be made, even before attempting to parse arguments to deduce the
> logging level or anything, and we need this trace to pop up *always*.
>
Why can't you initialise the logger and then start the logic assuming you
always use the logger ? Don't see the point.

> The logger is statically and implicitly pre-initialized (which I don't
> like => static initialization errors are difficult to debug and act upon),
> but the actual level and stream are set during the body of the program. I
> don't want to depend on any implicit initialization of the logger, because
> root permissions is the very first thing we need to check, even before
> configuring the logger.
>
As I said before. CRITICALs shouldn't be swallowed anywhere.

>  Something like this is what I would expect:
>
> //check root (and use fprintf(stderr, "") in purpose)
> //Parse arguments
> //Initialize logger
> //here **use** the logger, not before with unknown(implicit) initialization
>
>
1) initialise logger (logging level must not go under a certain level.
CRITICALs have not to be swallowed)
2) Do whatever you want assuming that logging works as it is supposed to
(otherwise it is a bug).

> In any case, this commit is merged. This review is late.
>
Backtracking is always possible and reviews are never late.


> —
> <https://github.com/IRATI/stack/pull/515#issuecomment-90878200>.
>



-- 
Francesco Salvestrini
neXtworks s.r.l.
Via Livornese, 1027
56122 S. Piero a Grado (PI) - ITALY
Phone: +39 050 3871647 / Fax: +39 050 3871601

msune:
2015-04-08 13:05 GMT+02:00 Francesco Salvestrini <notifications@github.com>:

> 2015-04-08 12:49 GMT+02:00 Marc Suñé <notifications@github.com>:
>
> > Actually, to properly clean resources exit() should be preferred instead
> > of return, so the other returns should be the ones to be changed. That's
> > not messing up, that's improving ;).
> >
> No, it is messing. a fallthrough of returns and one exit. Using returns,
> destructors (locally scoped) will be called. Using exit(), they won't.
> That's why there is (was) 1 single exit at most.
>

There are no local scoped variables because the check is the first thing
done in the main of both executables, so I don't get you. `return` does not
call `exit()` syscall, hence it is not freeing OS local resources straight
ahead, but you leave the OS to wipe it out for you (gc)



>
> > For the fprintf(), as you say, it *should*, but I don't want to depend on
> > the logger implicit initialization. This check is the very first one that
> > has to be made, even before attempting to parse arguments to deduce the
> > logging level or anything, and we need this trace to pop up *always*.
> >
> Why can't you initialise the logger and then start the logic assuming you
> always use the logger ? Don't see the point.
>

The sooner that you detect non-root and exit the best, because you prevent
the user to get _other_ errors (like logger initialization ones) before
this one. The non-root can induce errors (e.g. trying to write the logging
into a root owned file), and that's why the scope of checking should go
from higher level (are you root or not) to lower (e.g. argument parsing,
logger init).


> The logger is statically and implicitly pre-initialized (which I don't
> > like => static initialization errors are difficult to debug and act
> upon),
> > but the actual level and stream are set during the body of the program. I
> > don't want to depend on any implicit initialization of the logger,
> because
> > root permissions is the very first thing we need to check, even before
> > configuring the logger.
> >
> As I said before. CRITICALs shouldn't be swallowed anywhere.


> > Something like this is what I would expect:
> >
> > //check root (and use fprintf(stderr, "") in purpose)
> > //Parse arguments
> > //Initialize logger
> > //here **use** the logger, not before with unknown(implicit)
> initialization
> >
> >
> 1) initialise logger (logging level must not go under a certain level.
> CRITICALs have not to be swallowed)
>

Just parsing the arguments and initializing the logger can throw errors
that will pop up **before** the root check. As I said, errors should be
checked in the order that its scope requires. Not doing so confuses the
user.

salvestrini:
2015-04-08 13:24 GMT+02:00 Marc Suñé <notifications@github.com>:

> 2015-04-08 13:05 GMT+02:00 Francesco Salvestrini <notifications@github.com
> >:
>
> > 2015-04-08 12:49 GMT+02:00 Marc Suñé <notifications@github.com>:
> >
> > > Actually, to properly clean resources exit() should be preferred
> instead
> > > of return, so the other returns should be the ones to be changed.
> That's
> > > not messing up, that's improving ;).
> > >
> > No, it is messing. a fallthrough of returns and one exit. Using returns,
> > destructors (locally scoped) will be called. Using exit(), they won't.
> > That's why there is (was) 1 single exit at most.
> >
>
> There are no local scoped variables because the check is the first thing
> done in the main of both executables,


Was referring to your:

"... of return, so the other returns should be the ones to be changed.
That's not messing up, that's improving ;)."

It's introducing bugs (due to possibly "harshly interrupted" workflows).

That is: I don't agree as a general approach.


> so I don't get you. `return` does not
> call `exit()` syscall,


In C++, a return statement in main has the effect of leaving the main
function (destroying any objects with automatic storage duration) and
calling std::exit with the return value as the argument.

Just using "return(s)" is enough. Let *all* the destructors be executed.


> hence it is not freeing OS local resources straight
> ahead, but you leave the OS to wipe it out for you (gc)
>

I'm not compromising normal C++ behaviors this way. exit() here and there
is just breaking the workflows.

> > For the fprintf(), as you say, it *should*, but I don't want to depend
> on
> > > the logger implicit initialization. This check is the very first one
> that
> > > has to be made, even before attempting to parse arguments to deduce the
> > > logging level or anything, and we need this trace to pop up *always*.
> > >
> > Why can't you initialise the logger and then start the logic assuming you
> > always use the logger ? Don't see the point.
> >
>
> The sooner that you detect non-root and exit the best, because you prevent
> the user to get _other_ errors (like logger initialization ones)


However don't see why assuming a "sane" logger (which a fallback to
console) is not an option.

before
> this one. The non-root can induce errors (e.g. trying to write the logging
> into a root owned file),


Fallback to console.


> and that's why the scope of checking should go
> from higher level (are you root or not) to lower (e.g. argument parsing,
> logger init).



> > The logger is statically and implicitly pre-initialized (which I don't
> > > like => static initialization errors are difficult to debug and act
> > upon),
> > > but the actual level and stream are set during the body of the
> program. I
> > > don't want to depend on any implicit initialization of the logger,
> > because
> > > root permissions is the very first thing we need to check, even before
> > > configuring the logger.
> > >
> > As I said before. CRITICALs shouldn't be swallowed anywhere.
>

must not


> > > Something like this is what I would expect:
> > >
> > > //check root (and use fprintf(stderr, "") in purpose)
> > > //Parse arguments
> > > //Initialize logger
> > > //here **use** the logger, not before with unknown(implicit)
> > initialization
> > >
> > >
> > 1) initialise logger (logging level must not go under a certain level.
> > CRITICALs have not to be swallowed)
> >
>
> Just parsing the arguments and initializing the logger can throw errors
> that will pop up **before** the root check.


parse arguments and spit errors (wouldn't they be in the console this
time?). If the arguments are wrong, don't start...


> As I said, errors should be
> checked in the order that its scope requires. Not doing so confuses the
> user.



> > 2) Do whatever you want assuming that logging works as it is supposed to
> > (otherwise it is a bug).
> >
> > > In any case, this commit is merged. This review is late.
> > >
> > Backtracking is always possible and reviews are never late.
> >
> >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <https://github.com/IRATI/stack/pull/515#issuecomment-90878200>.
> > >
> >
> >
> >
> > --
> > Francesco Salvestrini
> > neXtworks s.r.l.
> > Via Livornese, 1027
> > 56122 S. Piero a Grado (PI) - ITALY
> > Phone: +39 050 3871647 / Fax: +39 050 3871601
> >
> > —
> > <https://github.com/IRATI/stack/pull/515#issuecomment-90881234>.
>
> >
>
> —
> <https://github.com/IRATI/stack/pull/515#issuecomment-90883675>.
>



-- 
Francesco Salvestrini
neXtworks s.r.l.
Via Livornese, 1027
56122 S. Piero a Grado (PI) - ITALY
Phone: +39 050 3871647 / Fax: +39 050 3871601

PR #518, opened 2015-03-29
Fix #517: cleanup of IPCP kernel state when detecting a crashed IPCP Daemon OS Process

### LINKED ISSUE #517: Cleanup IPCP kernel state when detecting that an IPCP Daemon has crashed
The IPCM should cleanup the IPCP kernel state when it detects that an IPCP Daemon has crashed (unless the IPCM is also shutting down).

edugrasa:
Sorry, forgot to create a separate branch to address the issue. The fix is tested.

### DIFF
--- a/rinad/src/ipcm/app-handlers.cc
+++ b/rinad/src/ipcm/app-handlers.cc
@@ -47,61 +47,65 @@ void IPCManager_::os_process_finalized_handler(
 			<< "terminated" << endl;
 	FLUSH_LOG(INFO, ss);
 
-	//Prevent any insertion/deletion to happen
-	rina::ReadScopedLock readlock(ipcp_factory_.rwlock);
+	{
+		//Prevent any insertion/deletion to happen
+		rina::ReadScopedLock readlock(ipcp_factory_.rwlock);
 
-	// Look if the terminating application has allocated flows
-	// with some IPC processes
-	collect_flows_by_application(app_name, involved_flows);
-	unsigned short ipcp_id = 0;
-	for (list<rina::FlowInformation>::iterator fit = involved_flows.begin();
-			fit != involved_flows.end(); fit++) {
+		// Look if the terminating application has allocated flows
+		// with some IPC processes
+		collect_flows_by_application(app_name, involved_flows);
+		unsigned short ipcp_id = 0;
+		for (list<rina::FlowInformation>::iterator fit = involved_flows.begin();
+				fit != involved_flows.end(); fit++) {
 
-		IPCMIPCProcess *ipcp = select_ipcp_by_dif(fit->difName);
-		if (!ipcp) {
-			ss  << ": Cannot find the IPC process "
-					"that provides the flow with port-id " <<
-					fit->portId << endl;
-			FLUSH_LOG(ERR, ss);
-			continue;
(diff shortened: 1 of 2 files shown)


rinad/src/ipcm/app-handlers.cc
```diff
@@ -47,61 +47,68 @@ void IPCManager_::os_process_finalized_handler(
...
 
-		{
-			//Auto release the read lock
-			rina::ReadScopedLock readlock(ipcp->rwlock, false);
-			ipcp_id = ipcp->get_id();
-		}
+			{
+				//Auto release the read lock
+				rina::ReadScopedLock readlock(ipcp->rwlock, false);
+				ipcp_id = ipcp->get_id();
```
vmaffione: This lock idiom is not correct - in general. You assign the local variable under the lock, but then you use it outside the lock (line 77): when you use it, the ipcp->get_id() value may have already been changed under your feet. Is that what you want? I know that the ipcp id should not change, but then why you need a lock to access the ipcp id?

msune: As far as I can see, it is correct.

The call to `IPCManager->deallocate_flow()` cannot be done with the IPCP readlock acquired or it will dead-lock. If the IPCP ID has changed the call will fail and that's it (the destruction will anyway deallocate the flow).

This line:
`rina::ReadScopedLock readlock(ipcp->rwlock, false);` is NOT acquiring the lock, but rather uses RAII to unlock the previously acquired rwlock in `select_ipcp_by_dif()` (the second parameter, set to false, indicates do not lock).

edugrasa: Yes, it is correct.

vmaffione: Right. Now that I've really examined it, I get that you use the ipcp->rwlock also as a reference counter - you get a reference in select_ipcp_by_dif and release it with that line. It's ok, but isn't that a bit obscure for the code reader?
Maybe we need some wrapper functions with more explicit names, like ``ipcp->read_get(), ipcp->read_put(), ipcp->write_get(), ipcp->write_put()``. Otherwise it is obscure - at least for me.

vmaffione: Right. Now that I've really examined it, I get that you use the ipcp->rwlock also as a reference counter - you get a reference in select_ipcp_by_dif and release it with that line. It's ok, but isn't that a bit obscure for the code reader?
Maybe we need some wrapper functions with more explicit names, like ``ipcp->read_get(), ipcp->read_put(), ipcp->write_get(), ipcp->write_put()``. Otherwise it is obscure - at least for me.

edugrasa: Vincenzo, this is pull request is about a minor fix to improve the stability of the current IPCM. The former pull request (aka the monster) was using this design extensively (in almost all the functions). The code in this pull request just follows the same design that is already there. 

We can discuss if it needs to improved and be made less obscure, but this would be another IPCM refactor. Therefore I would just merge this humble pull request in, which will make the live of people using the stack a bit easier ;)

vmaffione: Yes, agree.

Issue #478, opened 2015-03-06
pristine-1.1 soft lockup when trying 100 parallel flows from the same IPC Process


edugrasa:
Ok, not the most common thing to do with the stack, but probably points to an underlying locking problem. The flows were correctly allocated, and data was flowing for a while, but after some time the soft lockup appeared (the soft lockup happened at the server side). Here is the log.

[ 1080.412032] BUG: soft lockup - CPU#0 stuck for 23s! [rina-echo-time:7065]
[ 1080.412032] Modules linked in: normal_ipcp rina_default_plugin shim_tcp_udp shim_eth_vlan rinarp arp826 rina_personality_default 8021q garp stp llc nfsd auth_rpcgss oid_registry nfs_acl nfs lockd fscache sunrpc psmouse i2c_piix4 pcspkr serio_raw hid_generic usbmouse usbhid hid floppy
[ 1080.412032] CPU: 0 PID: 7065 Comm: rina-echo-time Not tainted 3.15.0 #6
[ 1080.412032] Hardware name: Xen HVM domU, BIOS 4.0.1 05/03/2013
[ 1080.412032] task: ffff88003db21840 ti: ffff88002c114000 task.ti: ffff88002c114000
[ 1080.412032] RIP: 0010:[<ffffffff816cdfe0>]  [<ffffffff816cdfe0>] _raw_spin_lock+0x20/0x30
[ 1080.412032] RSP: 0018:ffff88002c117e38  EFLAGS: 00000297
[ 1080.412032] RAX: 00000000000008c0 RBX: ffffffff8100180f RCX: ffff880029f173e0
[ 1080.412032] RDX: 00000000000008c1 RSI: 0000000000000035 RDI: ffff88001b2f7ad8
[ 1080.412032] RBP: ffff88002c117e38 R08: ffffffffa019b4a0 R09: ffff88002c0040e0
[ 1080.412032] R10: 00000000f778cb30 R11: 0000000000000000 R12: 00000000000130c0
[ 1080.412032] R13: ffff88003fc130c0 R14: ffffffff81084228 R15: ffff88002c117da8
[ 1080.412032] FS:  0000000000000000(0000) GS:ffff88003fc00000(0063) knlGS:00000000db5e5b40
[ 1080.412032] CS:  0010 DS: 002b ES: 002b CR0: 000000008005003b
[ 1080.412032] CR2: 00000000f7714000 CR3: 000000002c08d000 CR4: 00000000000006f0
[...]

PR #535, opened 2015-04-08
Pristine 1.2 #478 Added the required enhancements to handle concurrent flows

miqueltarzan:
Normal-ipcp and efcp use of locking has been enhanced to allow concurrent flows. Other components have been modified accordingly when needed.


vmaffione:
Can you please briefly explain what are EFCP_ALLOCATED and EFCP_DEALLOCATED for?

miqueltarzan:
EFCP_ALLOCATED and EFCP_DEALLOCATED are used to identify when an instance of EFCP can be used and when can be destroyed. For example, write and receive operations cannot progress once the EFCP has been deallocated but not yet removed. The destruction of an EFCP instance is not an atomic operation since it requires the removal of several other components like the DTP, DTCP, timers, etc. Under heavy workloads, deallocated but not completely removed EFCPs were a major problem, since operations like write and receive were trying to access structures that weren't there anymore.

Issue #539, opened 2015-04-16
Create ApplicationProcess base class in librina-application and refactor IPC Process


edugrasa:
To capture common DAF things that should go in all application processes (immediate use, allow both IPC process and MAd to extend from it, as well as measurement applications that want to leverage the DAF pattern and use the RIB, CDAP, etc..)

Refactor IPC Process accordingly (extend from ApplicationProcess)

msune:
+1

PR #540, opened 2015-04-16
Pristine 1.2 #539 Initial refactor of IPCP 

edugrasa:
* Moved former librina-application to librina-ipc-api (this is what there was there)
* librina-application holds now the daf components to build programmable distributed applications (policy set base classes and base ApplicationProcess classes are there)
* IPCP extends from that

More stuff needs to be moved to librina-application, but for the moment we should merge since a number of classes have changed, to prevent conflicts later on.

### DIFF
--- a/librina/include/librina/Makefile.am
+++ b/librina/include/librina/Makefile.am
@@ -8,6 +8,7 @@ nobase_pkginclude_HEADERS =			\
 	librina.h				\
 	logs.h					\
 	common.h				\
+	ipc-api.h                               \
 	application.h				\
 	configuration.h				\
 	ipc-daemons.h				\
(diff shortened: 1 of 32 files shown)


vmaffione:
583 additions and 1,693 deletions

If it's just refactoring, it should not be the case that we lost 1000+ lines of code. Am I wrong or you didn't add some new files to the repo?

edugrasa:
See the first line
 * Moved former librina-application to librina-ipc-api (this is what there was there)

vmaffione:
I know, that's what we already agreed on.

What I'm pointing out here is that it seems that the librina-ipc-api (the source files) are missing in this Pull Request, hence it seems we lost 1000 lines of code.

PR #542, opened 2015-04-20
Fixed problems in the makefile of sm-passwd plugin and IPCM plugin-load

edugrasa:
Due to latest code refactors (IPCM and DAF classes).


vmaffione:
Ack

PR #543, opened 2015-04-20
Allow IPCM configuration file to specify policy-sets for userspace components

vmaffione:
Allow IPCM configuration file to specify the name of policy sets to be selected at assign-to-dif time, for userspace IPCP components.

Pull request also includes update to test configuration files


edugrasa:
Tested and works as expected (with and without changes in the config file)

Issue #544, opened 2015-04-20
Move plugin-related code within the plugins folder of rinad/ipcp


edugrasa:
move any ps-set related code within the "plugins" folder so, ipcp/plugins/default folder containing plucin.cc  flow-allocator-ps.cc (what is now default-flow-allocator-ps.cc) etc.

edugrasa:
@vmaffione this issue can be closed

PR #545, opened 2015-04-20
Pristine 1.2 #544 Moved default plugin code to the plugins/default folder

edugrasa:
Addressed issue #544 
Also adapted the sm-passwd plugin.
Removed unused ipcp/include folder.
Fixed some IPCP tests.
Adapted related makefile.am and configure.ac files


vmaffione:
Tested, seems to work --> ACK

PR #549, opened 2015-04-22
Pristine 1.2 #539 Moving DAF generic things from IPCP to librina

edugrasa:
Changes (more work for #539):
   * moved plugin management code out of IPCP to librina-application
   * moved IPCP internal events code to librina-internal-events, decoupled internal events management form RIB Daemon
   * moved common parts of IRM to librina (management of N-1 flows)


vmaffione:
Tested, works ok to me --> ACK

PR #551, opened 2015-04-26
rinad: encoders: remove unnecessary prefixes

vmaffione:
Make .proto definitions more uniform by removing unnecessary namespace prefix.


msune:
ack

PR #552, opened 2015-04-27
Add manifest support for plugins

vmaffione:
Introduce manifest file support for plugins
- when IPCP recevies plugin-load command, it first loads the manifest and then invokes the plugin init function, checking that the PsFactories that ask for registration match against the advertised ones
- add plugin_get_info() method to IPCM, that could be used from MA
- add plugin-get-info console command to list plugin contents
- moved jsonpp to librina, distributed as librinajsoncpp library

This PR evidence a need for change for the current plugin loading procedure (init() function), that will be addressed in future PR


edugrasa:
ACK, works as expected if LIBJSONCPP minor version is set to 0.6.0

vmaffione:
I'm not sure we are safe with the external dependency, since jsoncpp has two branches, and does not seem to be well maintained. Also see here

https://github.com/open-source-parsers/jsoncpp/tree/master

maybe that was the reason to embed the code?

edugrasa:
if so, I would create another folder, at the same level of librina, rinad, rina-tools and linux, containing third-party libraries from which we need to build the sources from whatever reason. Then lbrina, rinad or rina-tools can use it without having multiple copies of the source code

vmaffione:
This issue is also connected to another question: Do we want all rina applications to use the same format for plugin manifest file ? And also the same plugging mechnism ?

edugrasa:
Applications that choose to reuse our classes yes, there is no requirement for applications using RINA to re-use the DAF tooling we provide. 

The primary application is for the IPCP,  the Management Agent and the applications we develop to generate traffic and measure it after going through a DIF. Other apps can also reuse if they which, but they don't need to to it.

vmaffione:
Yes, that's true. But in the end we need plugins/policyset support mainly in the IPCP and MAD, which happens to be in the same package (rinad). Going for another package (rina-tp) is reasonable, if not too expensive in terms of development.

edugrasa:
librina just provides a default way of loading policies. Developers may or may not use it, no obligation. If it is closed within rinad, any application that wants to reuse it will not be able to do it. 

We may not do another package (rina-tg) within PRISTINE, but maybe in the future (or other people will).

msune:
**Patch**

Some questions:

* As far as I can see this is only for user-space plugins right? Shouldn't this approach, i.e. the APIs in the IPCM, be consistent for both user-space *and* kernel space?
* [Minor issue]: do we want json keys be CamelCase or all lowercase or lowercase with hyphens or even underscores? So:

```
 v = root["PolicySets"];
```
or:
```
 v = root["policy-sets"];
```
or:
```
 v = root["policy_sets"];
```

Most of the parameters use camelCase, but not all:

https://github.com/IRATI/stack/blob/pristine-1.2/tests/conf/ipcmanager.conf-normalovertcpudp1#L125-131

**APIS**

I am pretty inline with what Eduard says about the policies in user-space for application processes that are not IPCPs. As far as I see, and correct me if I am wrong here, I see 2 types of policies:

a) Policies that all APs have (e.g. security component)
b) Policies that only IPCPs have (e.g. routing, RMT, EFCP...)

Within b) one could distinguish:

b.1) user-space
b.2) kernel space

The key here is to figure out how the APIs should look like. To me it is clear that everything that falls into a) should be set via librina, and also b) but __maybe__ in a different way? I wouldn't think so right now. I like quite a bit how is handled here:

https://github.com/IRATI/stack/blob/pristine-1.2/rinad/src/ipcp/ipc-process.cc#L88-120

But if possible, not sure, I would unify the thing:

```
int librina::select_policy_set(const std::string& ps_name, const std::string& component /* or better an enum*/);
```

And keep the "magic" of what it needs to be done within librina (e.g. dlopen etc..).

That would mean roughly (dumping my state of mind roughly here):

1. librina on bootstrap requests to the ipcm "give me the list of available policies and their meta-information"
2. [on background] librina will keep listening for new policies available or unavailable (unloading plugins). Note that this implies that the IPCM has always the catalogue of the available policies, for ALL components.
3. If the application, either an AP or an IPCP requests librina to load a policyset, librina has the meta-data on where "it is", meaning it knows if it needs to dlopen something or eventually load a kernel module. If a  kernel module needs to be opened => the IPCP requests the IPCM to load it (it needs to keep the state for that), otherwise (user-space) dlopen. Note that I am assuming that the kernel policies (IPCP ones) may not only be changed via the MAd/IPCM, which I think is possible. If not this could be done a priori by the IPCM before instructing the IPCM to load policy X.
4. Do the necessary stuff (policy set factories)
5. Return to the AP code 

Ofc, as Eduard was saying, the AP (not IPCPs) can freely choose not to use some policy provided by the librina system, in which case we could have some custom APIs (user-space only):

```
int librina::load_custom_plugin(const std::string& path_to_plugin);
int librina::unload_custom_plugin(const std::string& path_to_plugin);

//At this point, librina already knows the "custom" plugin policy sets in addition from the ones exposed by the IPCM so the API is the same
int librina::select_policy_set(const std::string& ps_name, const std::string& component /* or better an enum*/);
```

Thoughts?

**JSON library**

Regarding the issue of the C++11 and libjson; if there can be problems with distros, then I am in favour of being compiled on-board, but only once. I agree with Eduard there.

vmaffione:
@msune 

on Patch

- the manifest is meant to be the same for both user-space and kernel-space plugin. This patch introduces also modification to the userspace plugin code to implement the check of the manifest against plugin contents. This check must be done by IPCP, as agreed. Kernel plugins are LKM to modprobe, so in that case the manifest check will be implemented in a different manner (not in this patch)

- we can use "camelCase", then since is used everywhere, that's ok. The snipped you pointed out does not follow the convention because those are the names of IPCP components: those strings are used inside ipcp implementation to match internal component names (that do not follow camelCase convention), so I would leave them as they are.


on JSON library:

- yes, compiling on board is better, but I would avoid adding another package just for JSON, in PRISTINE. Alternative solution is to demand the manifest parsing/format to the application, while librina just receives a std::list<rina::PsInfo> when plugin_load is invoked. In this way we can keep json sources inside rinad, and the manifest parsing code (it's just a function) can be used by both IPCP and MAD, which are the target for PRISTINE policies - afaik in PRISTINE we don't foresee example applications that use policies (since Manager is not included in IRATI/stack). What do you say?


on API:
- can we talk about that in a separate (ML) thread? lots of things here, I would avoid generating confusion here

msune:
>>This check must be done by IPCP, as agreed. Kernel plugins are LKM to modprobe, so in that case the manifest check will be implemented in a different manner (not in this patch)

The point is precisely that I don't think the manifest has to be read inma different way. How to load the plugin, yes. Can you explain how different you rhink it can be?

Json: librina-json, compiled on board, and distributed as a separate library. Let me know if you need help with autotools.

vmaffione:
Yes, the manifest parsing will be the same --> rina::plugin_get_info(), for both userspace plugins and kernelspace plugins, I am not stating something different. What I say is that while userpsace plugins are loaded by IPCP daemon (and are then validated against the manifest by the IPCP daemon itself), kernelspace plugins are currently loaded manually, using modprobe - I think we should extend plugin-load/unload functionalities so that the modules are loaded from the IPCM. Consequently, while userspace plugins "publish" their policy sets to the IPCP daemon, kernelspace plugins "publish" their policy sets to the kernel, so there is the need for an upcall notification (kernel->IPCM) to inform the IPCM that a new plugin has been loaded, passing the published policy sets, so that the IPCM can check those agains the manifest.

msune:
>> kernelspace plugins "publish" their policy sets to the kernel, so there is the need for an upcall notification (kernel->IPCM) to inform the IPCM that a new plugin has been loaded, passing the published policy sets, so that the IPCM can check those agains the manifest.

Kernel->ipcm that a new plugin has been loaded?? It is the IPCM who _anyway_ has to load the plugin (modprobe)! 

So effectively the IPCM has to maintain this information, as I was explaining in the API section in the first contribution, and why it is relevant here. There are basically 2 use-cases.

a) If the MAd/console whatever external says: DIF X needs policy Y, the IPCM should know if it is a kernel or user-space and instruct either the KIPCM or the IPCP to do so. 

b)  **if and only if**, the DIF itself could potentially change its own policies based on a decision for instance between all the IPCPs in a DIF, then the IPCP via librina should we able to also request that _from_ userpace

vmaffione:
>> Kernel->ipcm that a new plugin has been loaded?? It is the IPCM who anyway has to load the plugin (modprobe)!

Yes, that's what I've written in the previous sentence. IPCM will modprobe, but still the "publishing" of kernelspace policy sets happens in kernel-space, so how do you check the published stuff against the advertised stuff (in the manifest)? Two solutions:
1) as soon as the the policy sets are published (in kernel) the kernel notifies the IPCM about the name/component/version of the policy set, so that IPCM can check against the manifest (which is read  in userspace IPCM)
2) [horrible imho] before modprobing, the IPCM loads the manifest and pushes the contained list<PsInfo> to the kernel. When modprobing happens, the kernel can match the publishing policy sets against the list<PsInfo>

Yes, we agree on (a) and (b), I was not talking about that.

msune:
2) is none sense. 

1) Why so complicated? The KIPCM only has to notify that a name+component factory has been loaded. The rest of information is in the manifest which the IPCM has already read.

By the way, this needs to work in a similar way  between IPCP user-space and IPCM. The IPCP needs to inform about the loaded factories during dlopen to verify the loading of the plugin was successful.

All this before the actual IPCP/AP in general uses it.

vmaffione:
>> 2) is none sense. 
Right.

>> 1) Why so complicated? The KIPCM only has to notify that a name+component factory has been loaded. The rest of information is in the manifest which the IPCM has already read.

Yes, this is what I wrote (or at least my intention :) ). First IPCM loads the manifest, then modprobes, then receives kernel notifications (published policy sets), so that it can check against the manifest.
To me we are on the same line. To you?

>> By the way, this needs to work in a similar way between IPCP user-space and IPCM. The IPCP needs to inform about the loaded factories during dlopen to verify the loading of the plugin was successful.

It's almost like that, already. The IPCP daemon checks the published policy sets against the manifest. If something goes wrong the plugin_load() fails, and the failure is propagated back to the IPCM

edugrasa:
ACK, compilation and usual tests work

Issue #550, opened 2015-04-23
Add support in PDU Forwarding Table for different next_hop policies.


miqueltarzan:
The PDU Forwarding Table main function is to provide a list of next hops for PDUs to be delivered to. Several strategies can be used to choose these next hops, and thus support for plugin policies must be available.

edugrasa:
@vmaffione this issue can be closed, or you prefer doing it when the changes to the PDU Forwarding Function configuration (via Netlink) are finished?

vmaffione:
I think this is related to
https://github.com/IRATI/stack/pull/553
which is already merged. Miquel, can we close it?

miqueltarzan:
Yes, it can be closed

PR #553, opened 2015-04-27
Pristine 1.2 #550 Added PFT policy set and changed pft_nhop into a policy.

miqueltarzan:
The PFT has now a policy set with the next_hop policy introduced. This policy allows for different ways of selecting the next hop, based on the PCI info, instead of using only the qos_id and the destination address.


vmaffione:
Tests seems OK --> ACK

PR #562, opened 2015-04-30
Fix #560 : Stop using CDAP sessions when sending CDAP message to address

### LINKED ISSUE #560: IPCP: flow allocation logic
Here

https://github.com/IRATI/stack/blob/pristine-1.2/rinad/src/ipcp/flow-allocator.cc#L717-L744

flow allocation is ongoing (on the initiator side), we've already used the DFT to get the remote address from the destination application name. As of my understanding, what we should do here is find the CDAP session towards the neighbor that is the next hop towards the remote address - in general next-hop != remote_address.

But from the code it seems that a "random" (the first one in a list) cdap session (e.g. a port_id) is selected.. this will work only if you have one neigbhor - unless I'm missing something.

edugrasa:
Work on #560 
 
* Fixed RIB Daemon
* Fixed Flow Allocator

### DIFF
--- a/librina/include/librina/cdap.h
+++ b/librina/include/librina/cdap.h
@@ -343,7 +343,15 @@ public:
 			int invoke_id);
 	static CDAPMessage* getCancelReadResponseMessage(Flags flags,
 			int invoke_id, int result, const std::string &result_reason);
+	static CDAPMessage* getRequestMessage(Opcode opcode, char * filter, Flags flags,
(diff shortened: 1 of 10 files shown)


vmaffione:
Seems to work as expected in a three-node normal-over-shim configuration (a -- b -- c) --> ACK

edugrasa:
Let's merge then?

PR #564, opened 2015-04-30
IPCM: fix policy-set handlers

vmaffione:
Return values of policy-set handlers operation were not propagated up correctly (error introduced by IPCM refactor).


msune:
ACK.

But please, do not merge before ACK

PR #566, opened 2015-05-04
Pristine 1.2

edugrasa:
Fixed a couple of bugs:

* Fixed handling of object value in IRM RIB objects, before could lead to seg faults
* Fixed IPCM deadlocks when processing unregister-from-dif and register-at-dif operations
* Fixed bug in IPCM register-at-dif respone handler


msune:
ACK

PR #568, opened 2015-05-05
First version of the Management Agent daemon (MAd)

msune:
This pull request contains:

* CDAP/RIB v2 libraries (alpha)
* Management Agent daemon with limited functionality:
  * Create IPCP
  * Destroy IPCP
  * Query IPCP RIB (string)
* console and MAd are in sync.

To launch the mad:

```
./ipcm -c ../etc/mad.conf -l DBG -a "console, scripting, mad"
```


sandervrijders:
Seems some build files are included in this PR as well ( rina-tools/build/config.status for example)

vmaffione:
Yes, someone did git add rina-tools/build by mistake. Please revert these.

vmaffione:
From a testing point of view, it is an ACK for me.
Tests performed:
1) [regression] three nodes experiment (MA--MB--MC), trying to ping/perf MA from MC --> OK
2) MAD experiment --> run the ipcm with MAD enabled and run the test manager to see that manager is able to create and configure IPCP.

msune:
Well spotted.

Done

edugrasa:
ACK also for the tests.

rina-tools/src/common/Makefile.am
```diff
@@ -5,3 +5,16 @@
...
 
 SUBDIRS = tclap
+
+noinst_LTLIBRARIES = librinaapp.la
+
+librinaapp_la_CFLAGS   =
+librinaapp_la_CPPFLAGS =				\
+	$(LIBRINA_CFLAGS)			\
+	$(CPPFLAGS_EXTRA)			\
+	-std=c++11
```
sandervrijders: Didn't we agree to stay away from C++11?

rinad/src/manager/Makefile.am
```diff
@@ -0,0 +1,29 @@
...
+	-I$(srcdir)/../common
+
+COMMONLIBS =					\
+	$(builddir)/../common/librinad.la	\
+	$(LIBRINA_LIBS)
+
+manager_CFLAGS   =
+manager_CPPFLAGS =					\
+	$(COMMONCPPFLAGS)	\
+	-std=c++11
```
sandervrijders: Same comment as before. C++11 flag is added here.

msune:
I don't know, but I don't think they are necessary.

@bernatgaston : could you please change C++11 statements and issue me a pull req?

@sandervrijders thanks for the review comments, but please we need also testers.

bernatgaston:
Hi,

I understand c++ version limitation for the stack. However the things you are pointing out are rina-tools  and the manager, used only for testing (while waiting for the authentic manager written in java). This manager will be removed, and for the rina-tools. I don't think we have to be that strict with things that are very high level (applications) and that are going to be removed (manager)

msune:
What is the exact feature you need from C++11?

bernatgaston:
I'm using class thread

msune:
In librina we have:

https://github.com/IRATI/stack/blob/master/librina/include/librina/concurrency.h#L119

So we should be able to remove it easily. The question is, do we delay the integration of this already ACKed pull req until this is fixed, or do we do a new pull req to remove the c++11 stuff?

bernatgaston:
I will submit it as an issue to be fixed and merge it since as I said, it is already being used in pristine 1.2 so anyways, it has to be removed from there too

PR #571, opened 2015-05-08
Remove WANT_PARACHUTE and install SIGSEGV handler

msune:
Remove useless WANT_PARACHUTE knob and install always the SIGSEGV signal
handler.

### DIFF
--- a/rinad/src/ipcm/main.cc
+++ b/rinad/src/ipcm/main.cc
@@ -49,8 +49,6 @@ using namespace std;
 using namespace TCLAP;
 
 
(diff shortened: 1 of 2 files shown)


vmaffione:
Seems ok. Could you give it a short try, to make sure we actually don't normally catch SIGSEGV signals? (I know it's extremely unlikely)

msune:
No need. They've never been without defining `WANT_PARACHUTE`

vmaffione:
Yeah, I was thinking to another scenario, actually. But right, it's not needed in this case.

PR #572, opened 2015-05-08
Ignore SIGINT in IPC processes

msune:
When debugging the IPCM in gdb, if SIGINT is not ignored, a Ctrl+C in the
debugger signals the child process and, since it is not captured, it dies.

This patch sets SIGINT to be ignored by IPCPs.

This patch requires commit 698eb30198226ac4b3e3d264af1183a5206a7fd1 PR to be
integrated first.


msune:
self-NAK

Unintentionally removed SIGPIPE exit()'s

msune:
Now it should be OK

edugrasa:
ACK, works as expected

PR #573, opened 2015-05-08
Fix ipcm deadlock when no events are received

msune:
Fixing (self-introduced) bug during IPCM shutdown process. Condition needs to
be signaled even if librina is not producing ANY event.


vmaffione:
ACK

PR #578, opened 2015-05-11
Changes in the serialized object for better swog bindings

bernatgaston:
Fioxed a librina swigg issue for MAD-Manager integration

### DIFF
--- a/librina/include/librina/cdap_rib_structures.h
+++ b/librina/include/librina/cdap_rib_structures.h
@@ -29,7 +29,7 @@ namespace cdap_rib {
 
 typedef struct SerializedObject {
 	int size_;
-	char* message_;
+	void* message_;
 } ser_obj_t;
 
 typedef struct cdap_params {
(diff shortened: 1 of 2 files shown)


vmaffione:
Could you please explain what is the point in this change?

bernatgaston:
by default. SWIG binds char* c++ to String in Java. This binding, which is usually good, is not suitable for all situations. In the specific case of the serialized object, char* means an array of bytes which is byte[] in java and not String. However, its very difficult to change (and maintain) only this special type binding in this specific occurence using SWIGG. Instead of this, void* is directly binded to byte[] so using this solution fewer changes must be done.

mcrotty:
Bernat, there is a larger set of files affected by this change.  How submits the further pull request.

bernatgaston:
If we used void* instead of char* bindings are automatically done to byte[] and moreover, few files need to be changed. I would go for this option.

PR #582, opened 2015-05-11
Initial extension of the PduFwdEntry data structures to accomodate alternatives port-ids

vmaffione:
This introduces librina and kernel support for the extended data structures, including userspace and kernelspace netlink encoding/decoding.

struct pdu_ft_entry has been renamed since its name was confusing (not really the data structure used by PFT implementation, just the one used to carry the RMT <--> RA message).


miqueltarzan:
ACK. I have tested with 2 and 3 nodes, forwarding and PDUFT DUMP work. Since these were the features affected by the changes, the PR can be merged.

Issue #583, opened 2015-05-12
pristine-1.3 Support for configuring RMT policies via DIF configuration


edugrasa:
There is some work started on that, but it needs to be completed. Includes:

* Support for RMT configuration in the config file (if configuration is not present defaults will be used)
* Update Netlink messages if necessary
* Update the RMT's kernel code to instantiate the requested policies

sandervrijders:
@edugrasa What's the status of this? Maintainers: @miqueltarzan @lbergesio

edugrasa:
This is already merged in pristine-1.3

PR #584, opened 2015-05-12
RMT policy configuration support via config file added, work on #583

edugrasa:
* Updated IPCM config parser and NL processing in user-space
* Updated NL in the kernel, and assign to DIF operation


edugrasa:
Will complete it later on with the code to actually select the policies in the RMT.

PR #590, opened 2015-05-13
Allowed shim-eth-vlan to have untagged frames.

sandervrijders:
A VLAN id 0 is interpreted as "no VLAN, untagged frames".


sandervrijders:
Tested regression tests, shim-eth-vlan with and without VLANs. Did a simple test with a normal DIF on top and rina-echo-time.

vmaffione:
Modifications are trivial and do not affect the old usage --> ACK

Issue #593, opened 2015-05-13
Bug in EFCP/KFA destruction


lbergesio:
In last implementation when the efcp is not destroyed until pending operations are finished. Nevertheless its state is changed to DEALLOCATED. If another PDU arrives to the EFCP in this state and the pending operations finally arrived to 0 the unbinding op of the ipcp on top is called. If this is a KFA a wake_up is called from interrupt context and it crashed.

The unbinding op of the KFA must defer this work in a workqueue to process the task in process context.

LOG:

[  164.361202] rina-efcp(DBG): EFCP connection destroy called
[  164.361222] rina-syscalls(DBG): Exiting SYSC_deallocate_port syscall body
[  164.399903] rina-dtcp(DBG): DTCP received FC (CPU: 1)
[  164.405965] rina-dtcp(DBG): SEQN: 60 N/Ack: 0 SndRWE: 1030087219 SndLWE: 0 RcvRWE: 50 RcvLWE: 0 newRWE: 1030087219 newLWE: 0 myRWE: 2496068818 myLWE: 0
[  164.425841] rina-kfa(CRIT): Do not call kfa_flow_deallocate in IRQ context
[  164.443130] ------------[ cut here ]------------
[  164.447086] kernel BUG at net/rina/kfa.c:198!
[  164.447086] invalid opcode: 0000 [#1] SMP 
[  164.447086] Modules linked in: shim_eth_vlan rinarp arp826 nfsd auth_rpcgss oid_registry nfs_acl nfs lockd fscache sunrpc rina_default_plugin shim_tcp_udp normal_ipcp rina_personality_default 8021q garp stp llc psmouse i2c_piix4 serio_raw pcspkr hid_generic usbmouse usbhid hid floppy
[  164.447086] CPU: 1 PID: 0 Comm: swapper/1 Not tainted 3.15.0 #56
[  164.447086] Hardware name: Xen HVM domU, BIOS 4.0.1 05/03/2013
[  164.447086] task: f5bb8e40 ti: f5be6000 task.ti: f5be6000
[  164.447086] EIP: 0060:[<f81afff7>] EFLAGS: 00210296 CPU: 1
[  164.447086] EIP is at kfa_flow_deallocate+0x157/0x170 [rina_personality_default]
[  164.447086] EAX: 0000003e EBX: dd288270 ECX: 00001808 EDX: 00200046
[  164.447086] ESI: 00000000 EDI: 00000002 EBP: f5407dec ESP: f5407dd4
[...]

PR #594, opened 2015-05-13
Fix for bug #593  in pristin1.3

lbergesio:
This fixes the problem identified at https://github.com/IRATI/stack/issues/593 related with kfa deallocation of flows when triggered by the reception of a pdu in the underlaying EFCP. When it is not possible to destroy the kfa's flow due to pending operations, the port is set to DEALLOCATED but the deletion task and the awakening of sleeping read syscalls is deferred to a dedicated workqueue (process context).

### DIFF
--- a/linux/net/rina/kfa.c
+++ b/linux/net/rina/kfa.c
@@ -40,11 +40,12 @@
 #include "kfa-utils.h"
 
 struct kfa {
-        spinlock_t             lock;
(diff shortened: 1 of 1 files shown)


lbergesio:
@vmaffione please, once I give you the ok after the concurrent flow tests please merge this one in pristine1.2 since it is also affected

Issue #585, opened 2015-05-12
pristine-1.3 Improve config file for defining DIFs


edugrasa:
Improvements:

* Define DIFs as templates in separate files, within the /dif-templates folder relative to the main configuration files.
* When defining a DIF in the config file we only need to fill in the DIF name and the template name (different DIFs can have the same template)
* DIF configurations only need to define the changes with respect to a basic template (included in dif-templates/default.dif)
* All DIF configurations can be updated while the IPCM is running (no need to reboot)

sandervrijders:
I'd avoid "All DIF configurations can be updated while the IPCM is running (no need to reboot)". Better to just restart the IPCM (eventually a system daemon) imho.

edugrasa:
Let the user choose what to do ;) If you want to restart the IPCM do it, otherwise now you'll have an option.

sandervrijders:
Wait, are you talking about watching the files with inotify to reload them upon save? Or I may be misunderstanding what you are proposing.

edugrasa:
Yeap, so after changing a file you could create a DIF with a new configuration (without affecting the DIFs already instantiated, of course).

sandervrijders:
Ok, that sounds reasonable. However, changing DIF description files of already instantiated DIFs (IPCPs in the system that have already joined that DIF) seems dangerous.

edugrasa:
Yes, I'm just talking about the DIF config *templates* (which are used to create new DIFs).

Existing DIFs will already have its own configurations, to be updated via the Manager when needed.

PR #597, opened 2015-05-14
Pristine 1.3 #585 Improved config file DIF definition

edugrasa:
Improvements:

* Define DIFs as templates in separate files, in the same folder of the main configuration file (ipcmanager.conf).
* When defining a DIF in the main config file we only need to fill in the DIF name and the template name (different DIFs can have the same template)
* DIF configurations only need to define the changes with respect to a basic template (called default.dif)
* All DIF configurations can be updated while the IPCM is running (no need to reboot)
* Added a show-dif-templates command to the CLI, to be able to visualize the templates loaded by the IPC Manager
* Updated all config files in tests/conf


lbergesio:
ACKed

Issue #595, opened 2015-05-13
app reserved ports do not unbind


bernatgaston:
When an app request a port for a flow allocation and the allocations does not work, it tries to unbind the port but it is not allowed to, so in the next try the port will be another one consuming ports without freeing them

[12611.422974] rina-personality-default(DBG): Calling wrapped function
[12611.436206] rina-pidm(DBG): The pidm bitmap find returned id 83 (bad = -1)
[12611.440031] rina-pidm(DBG): Bitmap allocation completed successfully (id = 83)
[12611.460170] rina-kfa(DBG): Flow pre-bound to port-id 84
[12611.466283] rina-ipcp-utils(DBG): Name at f4044f90 finalized successfully
[12611.473392] rina-ipcp-utils(DBG): Name at f4044f90 destroyed successfully
[12612.481001] rina-syscalls(DBG): Handling personality hook deallocate_port
[12612.488119] rina-syscalls(DBG): Calling personality hook deallocate_port
[12612.496718] rina-personality-default(DBG): Calling wrapped function
[12612.507324] rina-normal-ipc(ERR): Could not find flow 84 to deallocate
[12612.514927] rina-kipcm(ERR): Failed deallocate flow request for port id: 84
[12612.996570] rina-ipcp-utils(DBG): Name at f4044f90 finalized successfully
[12613.000670] rina-syscalls(DBG): Handling personality hook allocate_port
[12613.006221] rina-syscalls(DBG): Calling personality hook allocate_port
[12613.009754] rina-personality-default(DBG): Calling wrapped function
[12613.016996] rina-pidm(DBG): The pidm bitmap find returned id 84 (bad = -1)
[12613.020682] rina-pidm(DBG): Bitmap allocation completed successfully (id = 84)
[12613.037278] rina-kfa(DBG): Flow pre-bound to port-id 85
[12613.042452] rina-ipcp-utils(DBG): Name at f4044f90 finalized successfully
[12613.060818] rina-ipcp-utils(DBG): Name at f4044f90 destroyed successfully
[12614.071698] rina-syscalls(DBG): Handling personality hook deallocate_port
[12614.079946] rina-syscalls(DBG): Calling personality hook deallocate_port
[12614.090845] rina-personality-default(DBG): Calling wrapped function
[12614.103466] rina-normal-ipc(ERR): Could not find flow 85 to deallocate
[12614.112059] rina-kipcm(ERR): Failed deallocate flow request for port id: 85
[...]

PR #603, opened 2015-05-19
Fixed bug on not releasing port_ids after failed flow allocation resulting in port_ids exhaustion

lbergesio:
Related to #595 some other bugs where discovered and fixed by means of a new op in the normal_ipcp api. It has to be merged in pristine-1.2 and ported to pristine-1.3.

Tested and working, @bernatgaston  to test again and ack it.


bernatgaston:
Ack. It works, thanks!

PR #604, opened 2015-05-19
Addition of component-ps-common.[ch] files with exported functions to be reutilized by custom policy sets

### LINKED ISSUE #588: pristine1-3: Move common functionality of the default policy sets to the the stack internals so that common policies can be reutilized
In order to allow custom policy sets to take advantange of common policies of the default policy set without having to copy-pase or reimplement the code, the common functions will be moved to a <component-ps-common>.c file exporting them. Custom policy sets will only have to wrap these functions. This is a workaround before changing the granularity of the policy set.

lbergesio:
Fixes #588. Workaround to reutilize default code in custom policy sets. Still a different sdk design is needed for second phase.


vmaffione:
Modifications seem ok, tests work --> ACK

linux/net/rina/dtcp-ps-common.h
```diff
@@ -0,0 +1,63 @@
+/*
+ * Common implementation of a default policy set for DTCP
+ *
+ *    Leonardo Bergesio <leonardo.bergesio@i2cat.net>
```
vmaffione: Author should match the ones in  linux/net/rina/dtcp-ps-common.c .

linux/net/rina/dtp-ps-common.h
```diff
@@ -0,0 +1,49 @@
+/*
+ * Common function for DTP's policy set
+ *
+ *    Leonardo Bergesio <leonardo.bergesio@i2cat.net>
```
vmaffione: Author should match the ones in dtp-ps-common.c

linux/net/rina/rmt-ps-common.c
```diff
@@ -0,0 +1,217 @@
+/*
+ * Common parts of the common policy set for RMT.
+ * This is envisioned for reurilization in other custom policy sets
+ *
+ *    Leonardo Bergesio <leonardo.bergesio@i2cat.net>
```
vmaffione: Please keep the authors of the previous rmt-ps-default.c implementation (v.maffione@nextworks.it)

PR #607, opened 2015-05-20
Implementation of RTT estimator policy of DTCP in the common code

### LINKED ISSUE #592: pristine-1.3: Implement DTCP's RTT estimator policy
The RTT estimator policy of https://github.com/IRATI/stack/issues/591 will be the default RTT estimator policy, so it has to be implemented in the default (common) policy set

lbergesio:
Closes #592 and will be used as basis for RTT estimator policy of programmable congestion control policy


edugrasa:
ACK, tested and works

vmaffione:
Can you detail what estimator algorithm has been used?
When is the algorithm intended to kick in? (flow control or/and rtx control ?)
How is one supposed to test it from userspace (what configuration)?

lbergesio:
The algorithm used is detailed in "Algorithm2" schema in https://wiki.ict-pristine.eu/wp3/d32/D32-ProgrammableCC.

It is basically the one used in TCP but considering RINA's A timer. The implementation can be seen in int common_rtt_estimator(struct dtcp_ps * ps, seq_num_t sn) in dtcp-ps-common.c.

It is only used to calculate TR when there is retransmission control. EFCP specs were wrong about this point and were fixed after discussion with John. 

To test it you just need to create a flow with rtx control (-g 0 in rina-echo-time). Enable logs in the kernel, you will see a message saying something like "New TR calculated XXXms". Nothing different should be seen if not rtx control is used.

PR #609, opened 2015-05-21
Adding missing clean up functions in the n1_port creation when binding an upper normal IPCP

### LINKED ISSUE #608: Possible memory leak in rmt-ps-default
Branch: integration-1.2

The default rmt-ps, during port creation operations, does not correctly releases the reserved resources in case of error. Before linux/net/rina/rmt-ps-default.c, line 104, "return -1" command it shall be inserted a rmt_qgroup_destroy() order of the previously created queue group.

This shall be enough to fix it.

lbergesio:
fix #608


lbergesio:
Possible bug discovered, wait for ACK to merge

lbergesio:
@vmaffione merge this one, it is an easy one. The bug I mentioned may exist but will be tested and if needed pushed in an independent pull request

PR #614, opened 2015-05-26
Added ECN flag to the PDU flags and removed wrong flags

miqueltarzan:
The ECN flag was missing, while there were introduced flags specified for the SDU delimiting that had to be removed.


vmaffione:
ACK

Issue #612, opened 2015-05-25
Bug when creating an efcp connection


bernatgaston:
There is a bug when you try to create an efcp connection at app level, local flows.

May 25 14:38:12 irati-vm-bernat kernel: [   84.160624] rina-syscalls(DBG): Handling personality hook deallocate_port
May 25 14:38:12 irati-vm-bernat kernel: [   84.167817] rina-syscalls(DBG): Calling personality hook deallocate_port
May 25 14:38:12 irati-vm-bernat kernel: [   84.171920] rina-personality-default(DBG): Calling wrapped function
May 25 14:38:12 irati-vm-bernat kernel: [   84.176877] rina-kfa(DBG): Destroying kfa flow now...
May 25 14:38:13 irati-vm-bernat kernel: [   84.180865] rina-kfa(DBG): We are destroying flow 1
May 25 14:38:13 irati-vm-bernat kernel: [   84.180865] rina-kfa(WARN): Instance f0bd5100 SDU-ready FIFO is NULL
May 25 14:38:13 irati-vm-bernat kernel: [   84.180865] rina-pidm(DBG): Bitmap release completed successfully (port_id: 1)
May 25 14:38:13 irati-vm-bernat kernel: [   84.619440] rina-ipcp-utils(DBG): Name at f0e45980 finalized successfully
May 25 14:38:13 irati-vm-bernat kernel: [   84.623893] rina-syscalls(DBG): Handling personality hook allocate_port
May 25 14:38:13 irati-vm-bernat kernel: [   84.628516] rina-syscalls(DBG): Calling personality hook allocate_port
May 25 14:38:13 irati-vm-bernat kernel: [   84.632617] rina-personality-default(DBG): Calling wrapped function
May 25 14:38:13 irati-vm-bernat kernel: [   84.636834] rina-pidm(DBG): The pidm bitmap find returned id 0 (bad = -1)
May 25 14:38:13 irati-vm-bernat kernel: [   84.640483] rina-pidm(DBG): Bitmap allocation completed successfully (id = 0)
May 25 14:38:13 irati-vm-bernat kernel: [   84.644418] rina-kfa(DBG): Flow pre-bound to port-id 1
May 25 14:38:13 irati-vm-bernat kernel: [   84.647802] rina-ipcp-utils(DBG): Name at f0e45980 finalized successfully
[...]

PR #615, opened 2015-05-27
Pristine 1.2 #612 Fixed bug in creation of EFCP instances

miqueltarzan:
Fixed bad locking in the normal ipcp in connection creation.


linux/net/rina/ipcps/normal-ipcp.c
```diff
@@ -263,8 +263,10 @@ cep_id_t connection_create_request(struct ipcp_instance_data * data,
         cep_entry->cep_id = cep_id;
 
         spin_lock(&data->lock);
+
         flow = find_flow(data, port_id);
         if (!flow) {
+                spin_unlock(&data->lock);
```
vmaffione: This seems double locking. Can you check it?

PR #617, opened 2015-05-27
Initial support for IPCP RIB delegation

vmaffione:
This changeset introduce the initial support by means of which the MAD can forward a CDAP request message to an IPC process daemon, and receive a response back.

Currently, there is no mechanism that the IPCP can use to send a CDAP response to something different than a CDAP session on a flow. A mechanism is needed to redirect the response towards the IPCM, and will be introduced in future Pull Requests.


edugrasa:
ACK, works!

PR #620, opened 2015-05-28
Policy sets for kernel components selectable in the ipcm conf file

### LINKED ISSUE #619: Pristine-1.3: Add PS configuration in the conf file for the kernel components
The ipcm conf file should allow to configure the policy sets to be used by the kernel components: RMT, PFT, DTP and DTCP. 

The PS LKM should be compiled and loaded before hand as the ipcps modules.

lbergesio:
RMT, PFT, DTP and DTCP's policy sets can be selected in the ipcm conf file by defining them in the shim template. Default values are "default" and loads the default plugin. The field version is not checked.

RMT conf example in default.dif conf file:
     "rmtConfiguration" : {
        "pftPolicySet" : {
          "name" : "default",
          "version" : "0"
        },
        "rmtPolicySet" : {
          "name" : "default",
          "version" : "0"
        }
     },

RMT and PFT policy sets are selected when the normal IPCP is created (since RMT is then created). DTP and DTCP's policy sets are selected when a connection is created (since dtp an dtcp are then created). Current limitation is that all efcp instances in the same ipcp will load the same PS. 

A PolicyConfig class (user space) or policy struct (kernel) is used as a workaround so that its name field contains the name of the policy set to be loaded instead of the name of a specific policy (i.e. transmission_control). This was done to reutilize existing models.

The LKM of the policy set must be compiled and loaded previously to launching the IPCM (like any other IPCP module (shims, normal)).

Fixes #619


vmaffione:
The specification of what policy-set to use (per-component) in the config files should go to the "policySets" stanza and not in the ad-hoc ones (e.g. "rmtConfiguration").

edugrasa:
Actually it should be the opposite. Policy sets are part of each component
confifuration, not something isolated. The policy sets item should
disappear in the short/mid term.

We can discuss on Tuesday, since today is a day off here ;)

Cheers,

Eduarf

vmaffione:
Tomorrow will be day off here ;)
However, main point here is to avoid duplicating mechanisms. Current mechanism for specifying policy set was (and is) "policySets", since it was discussed and integrated in the code. We can change it of course, if we think it's necessary, that's ok. It is not acceptable that some components are configured through "policySets" and others using adhoc "rmtConfiguration" or "efcpConfiguration". And I have not heard about any discussion about this change/decision was in progress.

If we want to switch, this PR needs to be augmented with ad-hoc configuration stanzas to replace
        "security-manager" : "default",
        "flow-allocator" : "default",
        "namespace-manager" : "default",
        "routing" : "link-state",
        "resource-allocator" : "default"

edugrasa:
Hi Vincezo,

Enjoy your day off :)

I agree that we need to improve our communication and do proper release
planning meetings, where we put all the features we want to implement for a
release into the table and discuss them apropriately :)

Almost all the DIF configuration its policies. Templates for DIF
configuration have been discussed within WP5:
https://wiki.ict-pristine.eu/wp5/Specification-template-for-a-DIF, in an
email thread with not a lot of contributors (I think it was only John
replying).

The policy-sets configuration item has been used in parallel with other
mechanisms to specify policies since day one (qos cubes configuration, data
transfer constants, routing configuration was already there). To me the
"policy-sets" configuration item is a temporary mechanism that is needed
while the other components have not yet a "definitive" configuration
infrastructure in place (I also was not aware of the discussion of the
policy-sets configuration item and only saw it when it was merged, and
thought it was ok for a temporary mechanism). For example, the work we have
been doing with Ondrej on authentication and SDU protection includes the
configuration of the security manager component, with a bunch of policies:
https://wiki.ict-pristine.eu/wp4/d42/d42-authentication#Configuration-of-the-security-manager
.

As we go in depth into more components we can replace slowly the
policy-sets configuration item (but we can't do it at once, only when we
have a clear idea on the particular component). We need configuration items
for the Flow Allocator, the NamespaceManager, the Resource Allocator,
routing is redundant (it was already there before the introduction of the
policy-sets thing, but has to be properly named and refactored to reuse the
PolicyConfig with PolicyParameters structure), etc. as discussed in the WP5
configuration management task (to be refined, etc).

Summing up, let's speak on Wednesday :)

Cheers,

Eduard

2015-06-01 13:00 GMT+02:00 Vincenzo Maffione <notifications@github.com>:

edugrasa:
Hi Vincenzo,

Another thought.. In fact isn't 'policy-sets' already something different?
"Policy-sets" is used at IPC Process creation time, not at IPC Process
configuration time, right?

Cheers,

Eduard

2015-06-01 21:43 GMT+02:00 Eduard Grasa Gras <eduard.grasa@i2cat.net>:

msune:
@edugrasa: as far as I understood after speaking with Francesco, and as it was clarified I think in D2.3, the policy sets are the minimum "set" or "group" of policies that can be loaded without falling into configuration inconsistencies, specifically when there are policies from the same component or between several components that depend on each other to perform its task (e.g. SDU protection requires authentication information and key material from the RIBDaemon). 

This means that when no dependencies exist for a policy (type), a policy-set should be strictly equal to 1 policy. 

The fact that the stack currently defines a policy-set per component is just circumstantial, because the exact dependencies were not known at that time. We have to find out what are these cross-relations and just define the policy-sets > 1 policy for those cases.

To your latest question; IMO the policy-sets should be the minimum atomic set of policies to be used when creating **and** configuring an IPCP.

edugrasa:
Hi Marc,

The current policy-sets definition (it just really is a set of functions)
has a number of issues that have to be fixed in the second iteration of the
SDK (identified while using the SDK, which is fantastic and the reason why
we have two iterations :)). We should have a full discussion about this
once first iteration is over.

In RINA (architecture) we have policies, which apply to a single component
(see D2.2 for the list of policies per component). These policies are
typically implemented with a set of functions (which is the current
policy-set definition in the IRATI implementation). So really, a policy-set
as it is now in IRATI is the implementation of a single policy (in the
architecture).

Cheers,

Eduard

2015-06-02 7:22 GMT+02:00 Marc Suñé <notifications@github.com>:

msune:
>>In RINA (architecture) we have policies, which apply to a single component
(see D2.2 for the list of policies per component). 

The point is granularity. IMO, as defined in the spec, or at least how I read it, a policy affects a certain mechanism of a component, not the component as a whole. To put an example in SDU protection, CompressionPolicy is independent from EncryptionPolicy, and to change one you don't have to necessarily change the other (D2.2 SDUP section).

The point of the policy-sets was (taken from D2.3, section 5.1):

*A policy set is the atomic unit that the RPI can load and instantiate. It can be defined as a group of policies that the RPI system can load and instantiate without reaching an inconsistent state.*

This **differs from the current implementation**. The current implementation of policy-sets is one per component, and this is an intermediate step to reach above's goal, which requires further study.

But so far, it has been identified that certain policies have to share state and collaborate to carry on its work. If you don't use the notion of a policy-set to load these _inter-dependent_ policies, and you do it manually and not atomically (and transactionally), you run the risk of disrupting the IPCP operation and even affect the rest of the peers in the DIF.

So, to me, either:
* The spec is revised to check why there are these inter-dependencies, and consider to merge in a single policy, in which case you would only load policies => I don't think this makes sense.. 
*  We have to keep the notion of the policy set, in which case:
  * Independent policies should be a policy-set on its own
  * Only dependent policies should form complex policy-sets

vmaffione:
Hi,

  I'm ok with the approach, but we still could immediately create a
configuration stanza for each component, that at least cointains policy-set
information. E.g.:

[...]
"flowAllocatorConfiguration" : {
      "policySet" : {
          "name" : "default",
          "version" : "0"
       }
},
"namespaceManagerConfiguration" : {
      "policySet" : {
          "name" : "default",
          "version" : "0"
       }
},
[...]

Cheers,
  Vincenzo


2015-06-01 21:44 GMT+02:00 Eduard Grasa <notifications@github.com>:

> Hi Vincezo,
>
> Enjoy your day off :)
>
> I agree that we need to improve our communication and do proper release
> planning meetings, where we put all the features we want to implement for a
> release into the table and discuss them apropriately :)
>
> Almost all the DIF configuration its policies. Templates for DIF
> configuration have been discussed within WP5:
> https://wiki.ict-pristine.eu/wp5/Specification-template-for-a-DIF, in an
> email thread with not a lot of contributors (I think it was only John
> replying).
>
> The policy-sets configuration item has been used in parallel with other
> mechanisms to specify policies since day one (qos cubes configuration, data
> transfer constants, routing configuration was already there). To me the
> "policy-sets" configuration item is a temporary mechanism that is needed
> while the other components have not yet a "definitive" configuration
> infrastructure in place (I also was not aware of the discussion of the
> policy-sets configuration item and only saw it when it was merged, and
> thought it was ok for a temporary mechanism). For example, the work we have
> been doing with Ondrej on authentication and SDU protection includes the
> configuration of the security manager component, with a bunch of policies:
>
> https://wiki.ict-pristine.eu/wp4/d42/d42-authentication#Configuration-of-the-security-manager
> .
>
> As we go in depth into more components we can replace slowly the
> policy-sets configuration item (but we can't do it at once, only when we
> have a clear idea on the particular component). We need configuration items
> for the Flow Allocator, the NamespaceManager, the Resource Allocator,
> routing is redundant (it was already there before the introduction of the
> policy-sets thing, but has to be properly named and refactored to reuse the
> PolicyConfig with PolicyParameters structure), etc. as discussed in the WP5
> configuration management task (to be refined, etc).
>
> Summing up, let's speak on Wednesday :)
>
> Cheers,
>
> Eduard
>
> 2015-06-01 13:00 GMT+02:00 Vincenzo Maffione <notifications@github.com>:
>
> > Tomorrow will be day off here ;)
> > However, main point here is to avoid duplicating mechanisms. Current
> > mechanism for specifying policy set was (and is) "policySets", since it
> was
> > discussed and integrated in the code. We can change it of course, if we
> > think it's necessary, that's ok. It is not acceptable that some
> components
> > are configured through "policySets" and others using adhoc
> > "rmtConfiguration" or "efcpConfiguration". And I have not heard about any
> > discussion about this change/decision was in progress.
> >
> > If we want to switch, this PR needs to be augmented with ad-hoc
> > configuration stanzas to replace
> > "security-manager" : "default",
> > "flow-allocator" : "default",
> > "namespace-manager" : "default",
> > "routing" : "link-state",
> > "resource-allocator" : "default"
> >
> > —
> > <https://github.com/IRATI/stack/pull/620#issuecomment-107399958>.
> >
>
> —
> <https://github.com/IRATI/stack/pull/620#issuecomment-107681957>.
>



-- 
Vincenzo Maffione

vmaffione:
2015-06-02 0:02 GMT+02:00 Eduard Grasa <notifications@github.com>:

> Hi Vincenzo,
>
> Another thought.. In fact isn't 'policy-sets' already something different?
> "Policy-sets" is used at IPC Process creation time, not at IPC Process
> configuration time, right?
>

No, "policy-sets" are used at assign-to-dif time (so config time), not at
IPCP creation time.

Vincenzo


>
> Cheers,
>
> Eduard
>
>
> 2015-06-01 21:43 GMT+02:00 Eduard Grasa Gras <eduard.grasa@i2cat.net>:
>
> > Hi Vincezo,
> >
> > Enjoy your day off :)
> >
> > I agree that we need to improve our communication and do proper release
> > planning meetings, where we put all the features we want to implement
> for a
> > release into the table and discuss them apropriately :)
> >
> > Almost all the DIF configuration its policies. Templates for DIF
> > configuration have been discussed within WP5:
> > https://wiki.ict-pristine.eu/wp5/Specification-template-for-a-DIF, in an
> > email thread with not a lot of contributors (I think it was only John
> > replying).
> >
> > The policy-sets configuration item has been used in parallel with other
> > mechanisms to specify policies since day one (qos cubes configuration,
> data
> > transfer constants, routing configuration was already there). To me the
> > "policy-sets" configuration item is a temporary mechanism that is needed
> > while the other components have not yet a "definitive" configuration
> > infrastructure in place (I also was not aware of the discussion of the
> > policy-sets configuration item and only saw it when it was merged, and
> > thought it was ok for a temporary mechanism). For example, the work we
> have
> > been doing with Ondrej on authentication and SDU protection includes the
> > configuration of the security manager component, with a bunch of
> policies:
> >
> https://wiki.ict-pristine.eu/wp4/d42/d42-authentication#Configuration-of-the-security-manager
> > .
> >
> > As we go in depth into more components we can replace slowly the
> > policy-sets configuration item (but we can't do it at once, only when we
> > have a clear idea on the particular component). We need configuration
> items
> > for the Flow Allocator, the NamespaceManager, the Resource Allocator,
> > routing is redundant (it was already there before the introduction of the
> > policy-sets thing, but has to be properly named and refactored to reuse
> the
> > PolicyConfig with PolicyParameters structure), etc. as discussed in the
> WP5
> > configuration management task (to be refined, etc).
> >
> > Summing up, let's speak on Wednesday :)
> >
> > Cheers,
> >
> > Eduard
> >
> > 2015-06-01 13:00 GMT+02:00 Vincenzo Maffione <notifications@github.com>:
> >
> >> Tomorrow will be day off here ;)
> >> However, main point here is to avoid duplicating mechanisms. Current
> >> mechanism for specifying policy set was (and is) "policySets", since it
> was
> >> discussed and integrated in the code. We can change it of course, if we
> >> think it's necessary, that's ok. It is not acceptable that some
> components
> >> are configured through "policySets" and others using adhoc
> >> "rmtConfiguration" or "efcpConfiguration". And I have not heard about
> any
> >> discussion about this change/decision was in progress.
> >>
> >> If we want to switch, this PR needs to be augmented with ad-hoc
> >> configuration stanzas to replace
> >> "security-manager" : "default",
> >> "flow-allocator" : "default",
> >> "namespace-manager" : "default",
> >> "routing" : "link-state",
> >> "resource-allocator" : "default"
> >>
> >> —
> >> Reply to this email directly or view it on GitHub
> >> <https://github.com/IRATI/stack/pull/620#issuecomment-107399958>.
> >>
> >
> >
>
> —
> <https://github.com/IRATI/stack/pull/620#issuecomment-107727581>.
>



-- 
Vincenzo Maffione

edugrasa:
Hi Vincenzo,

Agree, let's create the missing configuration stanzas then :)

Cheers,

Eduard

2015-06-03 10:15 GMT+02:00 Vincenzo Maffione <notifications@github.com>:

lbergesio:
I agree too

linux/net/rina/ipcp-instances.h
```diff
@@ -94,9 +94,19 @@ struct efcp_config {
...
+/* Represents the configuration of the RMT */
+struct rmt_config {
+	/* The PS name for the PDU Forwarding Table (Next hop) */
+	struct policy * pft_policy_set;
+
+	/* The PS name for the RMT */
+	struct policy * rmt_policy_set;
+};
+
 /* Represents a DIF configuration (policies, parameters, etc) */
```
vmaffione: What's the point of storing the policy names there? Remember that policy-set names are already stored in the per-component "available policy set list", as soon as they are published by plugins.

edugrasa: This is the DIF configuration that comes from user-space, and the policies
the RMT has to be configured with (when assign-to-dif is called). You can
have a number of RMT policy-sets published by plugins in the kernel (for
different DIFs, for example), this is the configuration of an RMT for a
particular DIF.



2015-06-03 11:58 GMT+02:00 Vincenzo Maffione <notifications@github.com>:

vmaffione: Yes, the question arises from the fact that I know how SDK (policy-sets and
plugins) is implemented :)
I get what this is, but the question remains the same: what's the point in
storing those strings there? We already have policy-set factories (with
names and factory methods) stored in the per-component policy-set lists. It
seems unused and redundant information.

2015-06-03 12:16 GMT+02:00 Eduard Grasa <notifications@github.com>:

> In linux/net/rina/ipcp-instances.h
> <https://github.com/IRATI/stack/pull/620#discussion_r31609938>:
>
> >          struct policy * unknown_flow;
> >  };
> >
> > +/* Represents the configuration of the RMT */
> > +struct rmt_config {
> > +	/* The PS name for the PDU Forwarding Table (Next hop) */
> > +	struct policy * pft_policy_set;
> > +
> > +	/* The PS name for the RMT */
> > +	struct policy * rmt_policy_set;
> > +};
> > +
> >  /* Represents a DIF configuration (policies, parameters, etc) */
>
> This is the DIF configuration that comes from user-space, and the policies
> the RMT has to be configured with (when assign-to-dif is called). You can
> have a number of RMT policy-sets published by plugins in the kernel (for
> different DIFs, for example), this is the configuration of an RMT for a
> particular DIF. 2015-06-03 11:58 GMT+02:00 Vincenzo Maffione <
> notifications@github.com>:
> … <#14db8ec745f67bc0_>
>  In linux/net/rina/ipcp-instances.h <
> https://github.com/IRATI/stack/pull/620#discussion_r31608763>: > struct
> policy * unknown_flow; > }; > > +/* Represents the configuration of the RMT
> */ > +struct rmt_config { > + /* The PS name for the PDU Forwarding Table
> (Next hop) */ > + struct policy * pft_policy_set; > + > + /* The PS name
> for the RMT */ > + struct policy * rmt_policy_set; > +}; > + > /*
> Represents a DIF configuration (policies, parameters, etc) */ What's the
> point of storing the policy names there? Remember that policy-set names are
> already stored in the per-component "available policy set list", as soon as
> they are published by plugins. — Reply to this email directly or view it on
> GitHub <https://github.com/IRATI/stack/pull/620/files#r31608763>.
>
> —
> <https://github.com/IRATI/stack/pull/620/files#r31609938>.
>



-- 
Vincenzo Maffione

lbergesio:
It is required in order to parse the netlink message coming from user space. Nevertheless I agree there is no need to keep it after setting the policy sets. I will fix that destroying this struct (which is created in the netlink layer) after setting the PS. Doing so, there is no need also to keep it in the rmt struct so I will remove that reference too.

linux/net/rina/ipcp-instances.h
lbergesio: It is required in order to parse the netlink message coming from user space. Nevertheless I agree there is no need to keep it after setting the policy sets. I will fix that destroying this struct (which is created in the netlink layer) after setting the PS. Doing so, there is no need also to keep it in the rmt struct so I will remove that reference too.

vmaffione:
Yes, that's what I mean. Of course you need to parse the policy-set names from netlink, but once used you can thrown them away.

edugrasa:
ACK, confirmed working (assign to DIF, enrollment, flow allocation/deallocation, IPCP creation and destruction)

vmaffione:
There are merge conflicts coming from merging of #628. Could you please solve them?

lbergesio:
done

linux/net/rina/connection.c
```diff
@@ -59,6 +59,13 @@ struct conn_policies * conn_policies_create(void)
                 return NULL;
         }
 
+        tmp->dtp_ps = policy_create();
```
vmaffione: As discussed previously, what's the point in having this tmp->dtp_ps (which are strings, basically)? It's never used and you have the same info in dtp->dtp_ps. Minor point: you use the same names for both struct dtp_ps and struct policy, and I think this is confusing.

linux/net/rina/dtcp-utils.c
```diff
@@ -346,6 +349,14 @@ static struct dtcp_config * dtcp_config_create_gfp(gfp_t flags)
           }
         */
 
+        tmp->dtcp_ps = policy_create();
```
vmaffione: Same as above. The general question is: what's the point in storing these "struct policy" object? They seem to be allocated on create, deallocated on destroy and never used. Why don't we just purge away this code?

lbergesio: Both are used when dtp and dtcp are created:

 710         dtp = dtp_create(tmp->dt,
 711                          container->rmt,
 712                          tmp,
 713                          policy_name(connection->policies_params->dtp_ps),
 714                          connection);

You could delete that specific member of the connection struct but it would be an isolated policy_detroy in the efcp_create function.... Moreover, the connection_destroy function still needs to consider that if dtp_ps is present then free it. If we do a policy_destroy in efcp_create when we destroy the connection the pointer connection->dtp_ps won't be null and we will be freeing it again.

vmaffione: Ok for this specific case, but.. what about all the other cases? For example tx_control field in struct window_fctrl_config? It seems completely useless (waste priceless kernel memory and complicates code). I know, it's a separate issue from this Pull Request.

vmaffione:
Tested, ok --> ACK

vmaffione:
tested, it works --> ACK

linux/net/rina/dtcp-utils.c
lbergesio: Yes, actually there is a FIXME for that:

 645 /* FIXME This function has not been ported yet to use the DTCP policy set
 646  * parameters in place of struct dtcp_config. This because the code
 647  * tries to access "connection parameters" that are defined as DTCP
 648  * policies (in RINA) even if DTCP is not present. This has to
 649  * be fixed, because the DTCP policy set can exist only if DTCP
 650  * is present.
 651  */
 
we have to check if all the data that is later use will be present somewhere else (ps, state vector, etc). If his is the case, we can free the config structs once the corresponding instances are created and the config parameters that are later neede are passed somewhere. As you said, this is another issue, which could be created to address later on.

PR #623, opened 2015-05-29
Pristine 1.3: Improvements and fixes related with n1_port usage

lbergesio:
N1_PORT_STATE_DEALLOCATED, pending_ops counter and improvements in locking introduced in the rmt_n1_port struct in order to deal with detected bugs (i.e. if an n-1 ipcp unbinded the n1 port maybe due to destruction of the n1 ipcp, while the n ipcp was writing).

Minor fixes in dbg messages writting seq nums as ints.

Important bug in the rmt send_worker (tasklet) in a do_while condition.


vmaffione:
Tested, it works --> ACK

PR #624, opened 2015-06-01
Improve support for IPCP RIB delegation

vmaffione:
This patch add some functionalities to the IPCP implementation - IPCPRIBDaemon::generateCDAPResponse() - to make it possible for RIB objects to reply to CDAP requests coming from the IPCM (rather from a neighbor).

As an example of this functionality, OperationalStatusRIBObject::remoteReadObject() has been implemented, so that the IPCM is able to read the operational status of a running IPCP process whrough CDAP.


vmaffione:
Added Promise support to remove that hardcoded "78".

edugrasa:
ACK, confirmed working!

PR #625, opened 2015-06-02
stop copying DIF templates to the <installation folder/etc> when building rinad

edugrasa:
Since it's quite annoying that it overwrites the DIF templates of the user

### DIFF
--- a/rinad/src/ipcm/dif-template-manager.cc
+++ b/rinad/src/ipcm/dif-template-manager.cc
@@ -324,6 +324,10 @@ void DIFTemplateManager::augment_dif_template(rinad::DIFTemplate * dif_template)
 		}
 	}
 
+	if (dif_template->smConfiguration.policy_set_.name_ == std::string()) {
+		dif_template->smConfiguration = default_template->smConfiguration;
+	}
+
 	if (dif_template->addressPrefixes.size() == 0) {
 		for (std::list<rinad::AddressPrefixConfiguration>::iterator it =
 				default_template->addressPrefixes.begin();


msune:
This is something we will have to re-add sooner or later for users. IMO we should find out how to distribute the file only if the files aren't there yet. One way could be using the dist-hook:

http://www.gnu.org/software/automake/manual/html_node/The-dist-Hook.html#The-dist-Hook

But doesn't look entirely clean. A configure.ac check seems not clean enough either, because the file should be checked at compile time and not at configure time.

I will have a look later today if there is a better alternative.

msune:
I took it from here:

https://lists.gnu.org/archive/html/automake/2002-04/msg00163.html

And I put a bit of steroids in it:

```shell
marc@dev:~/git/personal/autotools-skeleton/src$ git diff Makefile.am
diff --git a/src/Makefile.am b/src/Makefile.am
index c97908a..a657f1f 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -33,4 +33,17 @@ bin_PROGRAMS = my_program
 my_program_SOURCES = my_program.c
 my_program_LDADD = libmy_lib.la -lpthread -lrt
 
+sysconf_DATA = default.dif.orig default2.dif.orig
+
+install-data-local: install-sysconfDATA
+       @FILES="default.dif default2.dif";\
+       for FILE in $$FILES ;\
+       do \
+               if test -f $(DESTDIR)$(sysconfdir)/$$FILE ; then \
+                       echo "$@ will not overwrite existing $(DESTDIR)$(sysconfdir)/$$FILE" ; \
+               else \
+                       echo "$(INSTALL_DATA) $(DESTDIR)$(sysconfdir)/$$FILE.orig $(DESTDIR)$(sysconfdir)/$$FILE"; \
+                       $(INSTALL_DATA) $(DESTDIR)$(sysconfdir)/$$FILE.orig $(DESTDIR)$(sysconfdir)/$$FILE; \
+               fi ;\
+       done
```

So in the sources there should be these files `default.dif.orig` `default2.dif.orig` (mind `.orig`), and the `make install` will copy **always** the `.orig` (overwrite), but will never overwrite `default.dif` ...

Sources:
```
marc@dev:~/git/personal/autotools-skeleton/src$ ls .
default2.dif.orig  default.dif.orig  distributable_header.h  internal_header.h  lib1.c  Makefile.am  Makefile.in  my_program.c  subdir
```

Result:

```
marc@dev:~/git/personal/autotools-skeleton/build$ ls /usr/local/etc/
default2.dif  default2.dif.orig  default.dif  default.dif.orig
```

There might be a way to simplify having to define `FILES=` but I am too tired. I think this will do the trick.

msune:
An additional comment. Be very careful with the indentation. The part from `@FILES...` until `done` must be indented with a hard tab (all lines)

Issue #628, opened 2015-06-05
pristine-1.3, RMT scheduling problem causes transmitter hang


vmaffione:
When trying to do rina-echo-time tests in a two nodes scenario

   $ rina-echo-time -w 10 -c 1000

client (transmission side) randomly hangs after a while (during the transmission).
Kernel logs shows an infinite sequence like the following

````
[   82.173199] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[   82.175155] rina-rmt-ps-common(ERR): Could not dequeue scheduled pdu
[   82.176922] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[   82.178827] rina-rmt-ps-common(ERR): Could not dequeue scheduled pdu
[   82.180602] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[   82.182505] rina-rmt-ps-common(ERR): Could not dequeue scheduled pdu
[   82.184273] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[   82.186185] rina-rmt-ps-common(ERR): Could not dequeue scheduled pdu
[   82.187950] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[   82.189850] rina-rmt-ps-common(ERR): Could not dequeue scheduled pdu
[   82.191622] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[   82.193555] rina-rmt-ps-common(ERR): Could not dequeue scheduled pdu
[   82.195309] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[   82.197235] rina-rmt-ps-common(ERR): Could not dequeue scheduled pdu
[   82.198608] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[   82.200581] rina-rmt-ps-common(ERR): Could not dequeue scheduled pdu
[   82.202429] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[   82.204362] rina-rmt-ps-common(ERR): Could not dequeue scheduled pdu
[   82.206117] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[   82.208052] rina-rmt-ps-common(ERR): Could not dequeue scheduled pdu
[   82.209809] rina-rqueue(WARN): queue ffff88000d762120 is empty, can't head-pop
[...]

PR #629, opened 2015-06-05
rmt: sender_worker: workaround to avoid unlimited spinning

vmaffione:
This is a workaround for https://github.com/IRATI/stack/issues/628, useful to make pristine-1.3 usable again.


lbergesio:
This is not ok, probably even as a workaround: if the pdu is retrieved from the scheduling policy and then passed to n1_port_write, checking the pointer in the while condition may produce unexpected results since the pdu has been disowned and it is converted to an sdu within n1_port_write

vmaffione:
#628 already fixed by #630

PR #630, opened 2015-06-08
Pristine 1.3: Fix for issue #628

lbergesio:
This contains a fix for #628 and does not include the workaround of pull request #629. It also fixes another bug discovered when the n1_port is destroyed within the do_while loop in the send_worker.

It needs to be better tested, in the same environment where the error was discovered.


vmaffione:
Tested, it fixes #628.

PR #632, opened 2015-06-10
Support for non-blocking flows, refactor and improvement of apps

edugrasa:
Support for non-blocking flows:

* Non-blocking read and write SDU operations. In case of blocking flows
* Applications requesting the allocation of a non-blocking flow must do so by setting a boolean in the flow specification. Applications that register to a DIF can also control if the flows targeting that application will be blocking or non-blocking

Refactor of apps
* Minimized copy-pasted code in the manager, rina-echo-time and rina-cdap-echo apps
* Removed the need to use C++11 for compilation (used librina concurrency classes instead of C++11 ones)
* rina-cdap-echo uses non-blocking flows (to be able to test that feature, since it is a very simple application that is not affected by the use of blocking vs non-blocking flows).

Improvement of rina-echo-time application
* Improved the ping mode. Uses the non-blocking flows API, so that when the timeout expires if the client didn't get the echo reply, the SDU is considered lost (default value for the timeout is 2 seconds, configurable via the rina-echo-time script). It also reports more statistics: %sdus lost, min/max/average/stdev of measured RTT.


vmaffione:
I think adding a timeout to sdu_write and sdu_read is not a good idea. You are mimicking select(), but read and write are a different matter. You are doing this because we don't have file descriptor support. I think we should address this problem rather than using these workarounds. At least, this is my opinion. By the way you are also changing the system call signatures, which is something that should not be done without proper discussion.

sandervrijders:
Yeah, I agree. If we really want this feature, why not go the extra mile and do it the proper way? Adding file descriptor support seems a good idea.

edugrasa:
I agree that adding file descriptor support is the way to go, but it will take a long time before we do that (not after the review). Adding "file descriptor support" means:

* Plug the KFA into the file system API in the kernel
* Use all the POSIX syscalls except for open (not powerful enough to pass application names and the flow requirement), we would need to implement the 
* Update all the apps

This is not going to happen before the review.

The current read/write API is temporary anyway, we have to move to the POSIX syscalls. This timeout facilitates coding some apps meanwhile (like rina-echo-time reporting packet loss), and it's already done. So, what's the problem with merging it? What we have right now is also a workaround.

vmaffione:
I know it cannot be done before the review (actually there is more work than you describe). Also, we don't know when/if we will have time to add such support.
Point is that this timeout will "hide" the real problem - that we miss file descriptors. If we need a workaround for rina-echo-time reporting packet loss, let's just add a timer in rina-echo-time, without changing all the stack down to kernel-space.

dstaesse:
I agree with Vincenzo and Sander, we should avoid merging workarounds in the main branches of PRISTINE, or we may end up with one workaround on top of another and yet another. This code can be kept in Edu's repo and people needing it can always pull from there.

edugrasa:
What do you mean merging workarounds? The current API is a workaround already, in the sense that the definitive solution won't come until we go for POSIX/file descriptors. I'm just improving a bit the current solution, so that it is a bit more usable, that's all.

Whatever we do until we use file descriptors is temporary. The current solution is already implemented, works, doesn't introduce bugs and is completely optional (if you don't want the timeout, don't use it).

dstaesse:
The problem is mainly that this wasn't discussed beforehand.
At least the following should be fixed:
* remove the timeout from the sduread call
* undo your changes in rina-echo-time, the calculation of the variance and standard error are wrong. You shouldn't update your average_rtt every time you receive an sdu, you have to record every rtt and in the end take the average of your entire sample population.

edugrasa:
Dimitri, let's finish discussions on Monday. But I just upgraded rina-echo-time so that it provides the same output as ping  (we need it for the experiments). I thought it was an obvious update todo (in addition to avoid unnecssary dependencies of C++11).

Regarding the variance and standard deviation formula, it's an online calculation, so that you don't need to store all the values and calculate at the end: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance (see the section "Online algorithm").

Regarding the timeout, it is not clear yet what to do, since some people (at least Marc & me) seees it as something useful.

We can finish discussions Monday morning. 

As for removing the timeout, let's discuss on

dstaesse:
cool, didn't know that online algorithm. We should move the application to its own repo, deprecate the perf functions and rename it rina-ping. Keeping the RTTs can also add us more useful statistics (median, quartiles,...), I don't know if there are online algo's for those...

msune:
@dstaesse this has to be done as a whole. We have to split the repos after the stage1 is complete, and have the builder have git-subrepos to build the entire thing, as of now.

vmaffione:
This will require ACKs from:
- @lbergesio 
- @sandervrijders 
- @vmaffione
- @miqueltarzan 
- @bernatgaston 
- @dstaesse

vmaffione:
When trying to do perf tests, it repeatedly fails

````
 irati/local/bin/rina-echo-time -t perf -c 100000
2827(1435590981)#librina.logs (DBG): New log level: INFO
2827(1435590981)#librina.nl-manager (INFO): Netlink socket connected to local port 2827 
Flow allocation time = 2.7426 ms
2827(1435590981)#rina-echo-time (ERR): Try again, resource not ready

````

vmaffione:
Ok, the last comment was addressed correctly. However, I see a strange behaviour when repeatedly running rina-echo-time client in ping mode (1 packet). This is the server log:

````
irati/local/bin/rina-echo-time -l
25448(1435665261)#librina.logs (DBG): New log level: INFO
25448(1435665261)#librina.nl-manager (INFO): Netlink socket connected to local port 25448 
25448(1435665264)#rina-echo-app (INFO): New flow allocated [port-id = 4]
25448(1435665264)#rina-echo-app (INFO): Flow torn down remotely [port-id = 4]
25448(1435665267)#rina-echo-app (INFO): New flow allocated [port-id = 3]
25448(1435665267)#rina-echo-app (INFO): Flow torn down remotely [port-id = 3]
25448(1435665270)#rina-echo-app (INFO): New flow allocated [port-id = 5]
25448(1435665270)#rina-echo-app (INFO): Flow torn down remotely [port-id = 5]
25448(1435665271)#rina-echo-app (INFO): New flow allocated [port-id = 4]
25448(1435665271)#rina-echo-app (INFO): Flow torn down remotely [port-id = 4]
25448(1435665272)#rina-echo-app (INFO): New flow allocated [port-id = 6]
25448(1435665272)#rina-echo-app (INFO): Flow torn down remotely [port-id = 6]
25448(1435665273)#rina-echo-app (INFO): New flow allocated [port-id = 3]
25448(1435665273)#rina-echo-app (ERR): Unkown test type 
25448(1435665275)#rina-echo-app (INFO): Flow torn down remotely [port-id = 3]
25448(1435665290)#rina-echo-app (INFO): New flow allocated [port-id = 3]
25448(1435665290)#rina-echo-app (INFO): Flow torn down remotely [port-id = 3]
25448(1435665291)#rina-echo-app (INFO): New flow allocated [port-id = 4]
25448(1435665291)#rina-echo-app (INFO): Flow torn down remotely [port-id = 4]
25448(1435665291)#rina-echo-app (INFO): New flow allocated [port-id = 5]
25448(1435665291)#rina-echo-app (ERR): Unkown test type ping
25448(1435665293)#rina-echo-app (INFO): Flow torn down remotely [port-id = 5]
````

So apparently there is some race condition so that sometimes the test_type variable is not set correctly.

dstaesse:
We need to resolve this soon or the work in here might die. On the one hand, it does contain useful features (improved stats, non-blocking I/O). On the other hand, functions like Application.applicationRegister(bool blocking) don't make much sense. The Application class needs a major overhaul (the dif_name should not be in the constructor etc), I already had a stab at it in tgen...
(https://github.com/dstaesse/traffic-generator/blob/master/src/simple_ap.h) and would port the tools (rina-echo-time is to become rina-ping and rina-perf) to inherit from that base class.

edugrasa:
Well, I just refactored the apps to avoid copy-pasted code and to avoid C+11  dependencies, that's all. The race condition reported by @vmaffione must have been there before, since I didn't touch this part (setting of the test_type variable). I would merge, and then @dstaesse can continue from here (as he says, this is just intermediate work, we have to move on)

vmaffione:
@edugrasa, I agree that we have to move on (and we will), and it could also be the case that the problem have been there before, but I've measured a regression. This means that while tests were always successful (before this PR), now they are not, they fail randomly. It does not matter where the error comes from, if we have regression we have regression, and we cannot ignore it. It's ok for me to let this go in, but after the regression has been removed - and leaving further modifications to @dstaesse.

edugrasa:
@vmaffione, ok didn't realize it was a regression, I'll check the cause and try to fix it.

vmaffione:
Ok! Also debugging it. It seems memory corruption, unfortunately :(

dstaesse:
We need to keep echo-time working in the repo, but don't spend too much time on this as I will probably scrap most of the current codebase, as it was - and still is - a mess.
Vincenzo, what is the format of that test? I would like to perform the same thing on the refactored tgen to see if it happens there too.

vmaffione:
@edugrasa 

I'm quite sure I've found the problem - and if I'm not mistaken it's a very classical one. Unfortunately, fixing it requires some overhaul of the Thread class, and all of their derived classes - there are a lot - and all the places where they are instantiated.

I'll provide an example to shows the problem (the same happens in other places):
````
EchoTimeServerWorker::EchoTimeServerWorker(ThreadAttributes * threadAttributes,
                                           const std::string& t_type,
                                           int port,
                                           int deallocate_wait,
                                           int inter,
                                           unsigned int max_buf_size,
                                           Server * serv) : ServerWorker(threadAttributes, serv)
{
        test_type = t_type;
        port_id = port;
        dw = deallocate_wait;
        interval = inter;
        max_buffer_size = max_buf_size;
        last_task = 0;
}
````
This is the constructor of a class that manages a rina-echo-time client. It's derived from ServerWork, which derives from SimpleThread, which in turn derives from Thread. Because of inheritance, you are forced to invoke the constructor of the derived class (ServerWorker in this case) **before** any other initialization can happen - for instance before EchoTimeServerWorker::test_type is initialized. Unfortunately, the ServerWork() constructor runs a thread (do_simple_thread_work), so it can happen - and it does happen - that do_simple_thread_work (and so EchoTimeServerWorker::internal_run())  gets executed before the EchoTimeServerWorker constructor has completed. In other words there is a race condition between the constructor and the internal_run() method. This is consistent with what I see (test_type sometimes empty at the very beginning of EchoTimeServerWorker::internal_run()). And seeing the variable empty is the best can happen, a segfault is more than welcome in this situation.

The root of this problem is the fact that the Thread constructor directly calls pthread_create(). With this happening I think it's not possible for any derived class (e.g. EchoTimeServerWorker) to safely work with a "context" (e.g. members of EchoTimeServerWorker) passed to the derived class' constructor. There will always be a race condition between the new thread scheduled (pthread_create()) and the constructor of the derived class.
You have two solutions:
1) [probably better] Thread constructor does not call pthread_create(), but just stores the function pointer and the argument. You have provide a Thread::start() method to start it (by calling pthread_create).
2) Don't derive from SimpleThread, but just define classes (e.g. Server Work) that have a pointer to an instance of SimpleThread as a member, and create the SimpleThread instance in some ServerWork::start() method - not in the ServerWork constructor.

In other words you need to move away pthread_create() from the constructor of the "Worker"/"Thread" classes.

dstaesse:
I'll have a look at this, but most likely I will ditch those wrapper classes alltogether and stick to POSIX threads. I think the code for these basic tools should be as clear as possible to people not having worked with the stack before (as @msune mentioned: these would be the entry point in using librina api).

edugrasa:
@dstaesse we should fix the problem and continue using the wrappers imho (if you don't use the wrappers the solution is not going to be portable, that was one of the reasons for having the wrappers). 

I propose the following. Let's fix it (I'll do it in a little while), @vmaffione I'll go for the first option you propose. Then let's merge the code and @dstaesse weight the pros and cons and decide how to continue from here.

vmaffione:
@edugrasa The plan seems ok to me.

edugrasa:
@vmaffione the latest commit should fix the issue (tested and verified working on my system, normal over shim)

msune:
@vmaffione +1 for option 1)

>>@dstaesse we should fix the problem and continue using the wrappers imho (if you don't use the pthread wrappers the solution is not going to be portable, that was one of the reasons for having the wrappers).

What portability? POSIX is portable across all BSD and Linux OSs.

However, I don't understand either the argument of @dstaesse; why do you want to remove a high level (wrapper) thread abstraction like these classes? Why do they make any harm?

edugrasa:
@msune agree for OSs that support pthreads, maybe this is the far majority of OSs (what about Windows?) Hiding low-level libraries was the main point for the C++ wrappers in concurrency.h and .cc

msune:
I would like to repeat; I agree on keeping the wrappers. 

But if you would want to port librina to windows you would have to rewrite almost entirely librina. Most of it may be under the hood, but also in the librina headers therefore the APIs; there are includes to `pthread.h`, `sys/types.h`.

vmaffione:
It's good thing to discuss these portability problems, but it's probably better to do it in another thread, this starts to be too crowded :)
@edugrasa testing your modifications as soon as they are ready. But let's be careful to add the start call everywhere (that was the major overhaul).

edugrasa:
@vmaffione modifications are ready (see latest commit). I've triple checked, all start calls should be in place.

vmaffione:
Ok, the race condition has gone away as expected. For me it's an ACK, waiting for 
    @lbergesio
    @sandervrijders
    @miqueltarzan
    @bernatgaston
    @dstaesse
to merge.

miqueltarzan:
ACK

lbergesio:
ack

bernatgaston:
ac

dstaesse:
If the camel once gets his nose in the tent, his body will soon follow.
ACK

sandervrijders:
ACK

-------- Oorspronkelijk bericht --------
Van: Dimitri Staessens <notifications@github.com> 
Datum:01-07-2015  19:34  (GMT+01:00) 
Aan: IRATI/stack <stack@noreply.github.com> 
Cc: Sander Vrijders <sander.vrijders@intec.ugent.be> 
Onderwerp: Re: [stack] Support for non-blocking flows, refactor and improvement
  of apps (#632) 

If the camel once gets his nose in the tent, his body will soon follow.
ACK

PR #633, opened 2015-06-11
Pristine 1.3 authentication and SDU protection

edugrasa:
A long description of the changes in this pull request is provided as part of D4.2 in the FP7 PRISTINE project wiki, chapters 2 (Authentication) and 5 (SDU Protection).

WARNING: Librina now needs openssl support. To install it (on Debian): 
 * apt-get install openssl
 * apt-get install libssl-dev

Security configuration:

* Unified security configuration under the security manager component (authentication, access control and SDU protection policies - encryption, errorcheck and TTL). 
* CRC and TTL are no longer activated as kernel compilation options, but via the configuration file (see examples in tests/conf/security.dif and tests/conf/security2.dif)

Encryption:

* Prototyped initial encryption "policy" (it doesn't use the SDK yet, so it is either on or off). 

Authentication:

* Implemented "SSH2" authentication policy which uses the DH key exchange mechanism to negotiate a shared secret, generates encryption keys, configures encryption and both parties authenticate each other using a pair of public/private RSA keys.


vmaffione:
I see some "enable_encryption" and "disable_encryption" NL messages going around. Does this mean that userspace IPCP issues those message to kernel in order to enable/disable encryption for all flows supported by a specific IPCP? And consequently, it is not possible for two flows supported by the same IPCP to have different encryption strategies (e.g. encryption activated or not), right?

edugrasa:
SDU protection is per N-1 port. Enable encryption/decryption enables encryption/decryption or both on a N-1 port.

vmaffione:
Ok that's good. But an "encryption" policy should be enabled/disabled using the SDK (select-policy-set, for instance), and this is not done for now, right? It's ok as initial step, but we will need a future PR addressing this.

It should then be possible for the SDK to "address" a port, something like

>>>> select-policy-set 4 rmt.7 AES256-ps

where we indicate that we want to change the policy for port 7.

edugrasa:
As reported in D4.2 (authentication chapter), there is no SDK support yet for encryption, error check (TTL and alike) and TTL. It is either enabled or not (per N-1 port). Ondrej is already working on adding SDK support for all of this.

linux/net/rina/du-protection.h
```diff
@@ -32,4 +33,16 @@ bool    dup_ttl_set(struct pdu_ser * pdu, size_t value);
 ssize_t dup_ttl_decrement(struct pdu_ser * pdu);
 bool    dup_ttl_is_expired(struct pdu_ser * pdu);
 
+int dup_encrypt_data(const char              * src,
```
sandervrijders: Shouldn't this follow the same convention as the function signatures above? E.g. use pdu_ser instead of char *.

edugrasa: agree, fixing that ...

edugrasa: Fixed!

vmaffione:
Test configuration files added seem not to be right. I see:
- tests/conf/ipcmanager.conf-sduprotection1: but it does not include a "*.dif" file. Could you please let it include a template, so that we have uniform conf files?
- tests/conf/security.dif and tests conf/security2.dif: they should be an extension of default.dif, but I see only the security-related configuration specified.

edugrasa:
The first file was wrong, a leftover from older versions of the code.

All the security configuration is optional (except for the selection of the "policy-set" of the security manager, which is already in the configuration file). All DIF templates extend from default.dif, that's why security.dif and security2.dif only need to contain the modified security configuration.

vmaffione:
I don't get from the security.dif file whether CRC and TTL are enabled by default or not.. how does it work?

edugrasa:
There's a default Authorization/SDU protection profile (defining authentication and SDU Protection policies) and then you can set more specific ones for different N-1 DIFs. 

Inside each Authorization/SDU Protection profile you can choose a particular Error Check (CRC is one form of error check) policy and a particular TTL policy (right now there's a single hardcoded policy, but the configuration already supports multiple, so that we don't have to change it later on). So, for example:

              "TTLPolicy" : {
                 "name" : "default",
                 "version" : "1",
                 "parameters" : [ {
                    "name" : "initialValue",
                    "value" : "50"
                  } ]
                },
                "ErrorCheckPolicy" : {
                   "name" : "CRC32",
                   "version" : "1"
                }

If you don't want ErrorCheck or TTL, just don't add these configuration items within the AuthSDUProtection profile. For example:

        "authSDUProtProfiles" : {
           "default" : {
              "authPolicy" : {
              	"name" : "PSOC_authentication-none",
          	"version" : "1"
              }
           }
        }

This profile uses no ErrorCheck policy, no TTL, no encryption and the "authentication-none" authentication policy.

librina/include/librina/common.h
```diff
@@ -616,6 +623,21 @@ class SerializedObject {
         void initialize(const SerializedObject& other );
 };
 
+struct UcharArray {
```
vmaffione: This class seems the same as SerializedObject. Why?

vmaffione:
2015-06-18 22:32 GMT+02:00 Eduard Grasa <notifications@github.com>:

> There's a default Authorization/SDU protection profile (defining
> authentication and SDU Protection policies) and then you can set more
> specific ones for different N-1 DIFs.
>
Ok, but I don't get how this is specified in the configuration file. I see
the definition of one profile called "default" (is this the default one
that you are referring to?). But how is one supposed to specify a
"specific" profile (and the associated N-1 DIF) ?


> Inside each Authorization/SDU Protection profile you can choose a
> particular Error Check (CRC is one form of error check) policy and a
> particular TTL policy (right now there's a single hardcoded policy, but the
> configuration already supports multiple, so that we don't have to change it
> later on). So, for example:
>
>           "TTLPolicy" : {
>              "name" : "default",
>              "version" : "1",
>              "parameters" : [ {
>                 "name" : "initialValue",
>                 "value" : "50"
>               } ]
>             },
>             "ErrorCheckPolicy" : {
>                "name" : "CRC32",
>                "version" : "1"
>             }
>
> If you don't want ErrorCheck or TTL, just don't add these configuration
> items within the AuthSDUProtection profile. For example:
>
>     "authSDUProtProfiles" : {
>        "default" : {
>           "authPolicy" : {
>             "name" : "PSOC_authentication-none",
>         "version" : "1"
>           }
>        }
>     }
>
> This profile uses no ErrorCheck policy, no TTL, no encryption and the
> "authentication-none" authentication policy.
>
> —
> <https://github.com/IRATI/stack/pull/633#issuecomment-113281444>.
>



-- 
Vincenzo Maffione

librina/src/netlink-parsers.h
```diff
@@ -612,6 +611,39 @@ int putSecurityManagerConfigurationObject(nl_msg* netlinkMessage,
...
+    SAUTHP_AUTH_PROFILE,
+    __SAUTHP_ATTR_MAX,
+};
+
+#define SAUTHP_ATTR_MAX (__SAUTHP_ATTR_MAX -1)
+
+int putSpecificAuthSDUProtectionProfile(nl_msg* netlinkMessage,
+					const std::string& under_dif,
+			   	        const AuthSDUProtectionProfile& object);
+int parseSpecificSDUProtectionProfile(nlattr *nested,
```
vmaffione: Shouldn't this called parseListOf... ?

librina/include/librina/common.h
edugrasa: The problem is that serialized object is a char array, and I needed unsigned char array to interact with the openssl APIs. Maybe we can change all the serialized object stuff by UCharArray, but I wasn't sure and it was a lot of changes. Let's see if a SerializedObject (which is really a class wrapping a char array) can be updated to unsigned char array (in fact, to represent a raw byte the usual thing to use is an unsigned char, not a char, right?)

edugrasa:
You can take a look at https://github.com/edugrasa/stack/blob/pristine-1.3-sdu-protection/tests/conf/security2.dif for an example of this.

librina/src/netlink-parsers.h
edugrasa: No, this function parses a  single specific SDU protection profile and adds it to the map.

vmaffione:
Ok got it, thanks. Tests work OK to me - using either default.DIF and security.DIF. ACK.

PR #634, opened 2015-06-11
Allow kernel compilation with gcc 5.x

vmaffione:
The current kernel version (3.15) used by IRATI is old, and does not contain the header necessary to compile the kernel with gcc 5.x. The file contained in this pull request has been taken from linux 4.0 sources.

### DIFF
--- a/linux/include/linux/compiler-gcc5.h
+++ b/linux/include/linux/compiler-gcc5.h
@@ -1,0 +1,67 @@
+#ifndef __LINUX_COMPILER_H
+#error "Please don't include <linux/compiler-gcc5.h> directly, include <linux/compiler.h> instead."
+#endif
+
+#define __used				__attribute__((__used__))
+#define __must_check			__attribute__((warn_unused_result))
+#define __compiler_offsetof(a, b)	__builtin_offsetof(a, b)
+
+/* Mark functions as cold. gcc will assume any path leading to a call
+   to them will be unlikely.  This means a lot of manual unlikely()s
+   are unnecessary now for any paths leading to the usual suspects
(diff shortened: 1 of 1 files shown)


vmaffione:
Device drivers compilation fails at some point. I cannot really find the proper LTS patches for 3.15. Anyone has some suggestions?

sandervrijders:
Updating our kernel version is probably a no-go? How much effort does it require?

sandervrijders:
3.15 was never LT AFAIK. We forwarded to this kernel because we needed it for the shim HV.

msune:
I don't think gcc 5.1 is right now a priority for us, is it?

I would avoid to spend time on it, and spend it when we really need it; when we have to rebase the entire thing and propose it main-stream. Which is far from now.

vmaffione:
It's not a priority, and we are not spending time at all on it. It was just
a compilation trial in background :)
I was just asking if someone had some good idea on fastly solving this.

In the end I think the solution could be to switch to an LTS kernel version
- our 3.15 is not LTS. The closest one is 3.18, which is what I would try
to merge in -   to be clear, very very low priority and background.

2015-06-19 12:39 GMT+02:00 Marc Suñé <notifications@github.com>:

> I don't think gcc 5.1 is right now a priority for us, is it?
>
> I would avoid to spend time on it, and spend it when we really need it;
> when we have to rebase the entire thing and propose it main-stream. Which
> is far from now.
>
> —
> <https://github.com/IRATI/stack/pull/634#issuecomment-113467202>.
>



-- 
Vincenzo Maffione

vmaffione:
This is not working, and not going to be addressed soon. Will be moved to an issue

PR #670, opened 2015-07-08
kernel: pff-ps: add pff_port_state_change() op

vmaffione:
The pff_port_state_change() operation has been added to the policy set in order to receive notifications about flow going up or down.

### DIFF
--- a/linux/net/rina/ipcps/normal-ipcp.c
+++ b/linux/net/rina/ipcps/normal-ipcp.c
@@ -418,6 +418,10 @@ static int normal_nm1_flow_state_change(struct ipcp_instance_data * data,
 {
 	LOG_INFO("N-1 flow with pid %d went %s", pid, (up ? "up" : "down"));
 
+	if (rmt_pff_port_state_change(data->rmt, pid, up)) {
+		LOG_ERR("rmt_port_state_change() failed");
+	}
+
 	return 0;
 }
 
--- a/linux/net/rina/ipcps/shim-eth-vlan.c
+++ b/linux/net/rina/ipcps/shim-eth-vlan.c
@@ -1484,8 +1484,9 @@ static int ntfy_user_ipcp_on_if_state_change(struct ipcp_instance_data * data,
 			continue;
                 }
 
-		flow->user_ipcp->ops->nm1_flow_state_change(data, flow->port_id,
-							    up);
+		flow->user_ipcp->ops->
+                            nm1_flow_state_change(flow->user_ipcp->data,
+                                                  flow->port_id, up);
         }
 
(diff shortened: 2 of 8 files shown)


vmaffione:
Maintainers: @miqueltarzan @sandervrijders

miqueltarzan:
ACK

sandervrijders:
Seems to fail:

```
[  629.211076] rina-shim-eth-vlan(DBG): Ignoring event 9 on device eth23.110
[  629.211892] rina-shim-eth-vlan(INFO): Device eth23.110 goes down
[  629.211896] rina-normal-ipc(INFO): N-1 flow with pid 1 went down
[  629.211898] rina-normal-ipc(ERR): rmt_port_state_change() failed
```

sandervrijders:
As discussed with @vmaffione in chat: This seems due to memory corruption somewhere. Accessing 'ps' at different moments in time:

```
[  126.007228] rina-pff(DBG): nhop: Pointer value is ffff8803286bce40
[  140.163830] rina-pff(DBG): port state change: Pointer value is ffff880326c75b08
```

Looking for a solution and proposing a patch in a different PR once it is found.

sandervrijders:
@miqueltarzan @lbergesio @edugrasa 

Been investigating with @vmaffione the whole day and it seems we are facing an unknown memory corruption problem somewhere in the stack (kernel space). This PR sometimes works and sometimes it doesn't, although it doesn't do any memory operations itself. When I print out the pointers, it reveals they are sometimes corrupted over a certain timeframe. So, please help squash this bug ASAP, since it is very blocking. Creating an issue.

msune:
Plwhenever reporting a bug (or possible bug) please attach the steps to reproduce it, so that it can be confirmed and fixed.

sandervrijders:
Right, so the steps to reproduce:

1) enroll between two nodes
2) bring an interface down on either one of the nodes

(Maybe step 1) is not even needed, didn't try it)

edugrasa:
Question for @sandervrijders . The memory corruption you see is always in the ps pointer?
Or you've seen it in other places as well? (trying to find the bounds of the problem)

sandervrijders:
I've also seen it in the RMT pointer that is kept in normal-ipcp.c

vmaffione:
Confirmed, sometimes seeing it in the ps pointer, sometimes I see the RMT pointer zeroed (which cannot be).

vmaffione:
Bug should be fixed now (it is on my side), please confirm.

edugrasa:
ACK, problem fixed (I also didn't notice the mistake when reviewing the PR.. we need holidays :))

PR #673, opened 2015-07-09
Fix kernel plugins

vmaffione:
This patch fixes compilation problems with the non-default kernel plugins and adds support for the "make install" target.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -156,6 +156,11 @@ M: Leonardo Bergesio <leonardo.bergesio (at) i2cat.net>
 F: linux/net/rina/rnl*
 F: linux/net/rina/syscall*
 
+Kernelspace plugins:
+M: Leonardo Bergesio <leonardo.bergesio (at) i2cat.net>
+F: plugins/red/*
+F: plugins/dummy/*
+
 
 User space: librina
 -------------------
--- a/plugins/dummy/Makefile
+++ b/plugins/dummy/Makefile
@@ -21,8 +21,7 @@ clean:
 	rm -r -f *.o *.ko *.mod.c *.mod.o Module.symvers .*.cmd .tmp_versions modules.order
 
 install:
-	@echo "This target has not been implemented yet"
-	@exit 1
+	$(MAKE) -C $(KDIR) M=$$PWD modules_install
 
 uninstall:
 	@echo "This target has not been implemented yet"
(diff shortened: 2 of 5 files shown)


vmaffione:
Maintainers: @msune @lbergesio

msune:
I did not realise that a previous PR from @lbergesio  was adding plugins/ folder.

Why is this folder outside of the kernel and uses an adhoc build system? Wouldn't it make more sense to be under the linux/ folder being kernel modules?

vmaffione:
Yes this is an interesting point o discuss. The idea of that plugins dir is that the plugins contained there are "external", so potentially written by users --> they should be out-of-tree kernel-modules. From a packaging point of view, we deliver the OpenIRATI packages (kernel, librina, etc), but if an user wants to write a custom plugin, that plugin wouldn't be part of our kernel package, so it has to be a kernel module. Does this make sense to you?

lbergesio:
ack

msune:
I don't quite get it... Where are kernel modules generally compiled? IMO this is in-tree.

You are arguing to move them out of the tree because of authorship and packaging issues, but these plugins are written by us and you need "reference" or "supported" plugins.

My suggestion would be:
* Plugins that are "actively maintained" by irati community go in-tree
* Users can always compile out-of-tree plugins on their own, but the final target, if the plugins is open-source and wants to contributed to the community, is to reach the in-tree plugins/ folder.

vmaffione:
2015-07-09 11:01 GMT+02:00 Marc Suñé <notifications@github.com>:

> I don't quite get it... Where are kernel modules generally compiled? IMO
> this is in-tree.
>
It depends on what you need, out-of-tree compilation is way faster and is
the only option when it makes no sense for linux to include your code
upstream.


> You are arguing to move them out of the tree because of authorship and
> packaging issues, but these plugins are written by us and you need
> "reference" or "supported" plugins.
>

There is already the default plugin for that, which is in-tree.

> My suggestion would be:
>
>    - Plugins that are "actively maintained" by irati community go in-tree
>    - Users can always compile out-of-tree plugins on their own, but the
>    final target, if the plugins is open-source and wants to contributed to the
>    community, is to reach the in-tree plugins/ folder.
>
>
Yes but it's also handy to have some "example" of out-of-tree plugins (e.g.
dummy is for that), so that external user can use them as template.

>
>
> —
> <https://github.com/IRATI/stack/pull/673#issuecomment-119880722>.
>



-- 
Vincenzo Maffione

msune:
ok, got convienced to have plugins/ folder, but please let's move RED to in-tree as soon as it is finished.

Therefore ACK

PR #674, opened 2015-07-09
rinad: ipcm: implement plugin_load_kernel()

vmaffione:
This patch allows ipcm to modprobe/rmmod plugin modules.

Maintainers: @msune

### DIFF
--- a/rinad/src/ipcm/ipcm.cc
+++ b/rinad/src/ipcm/ipcm.cc
@@ -1157,7 +1157,7 @@ IPCManager_::select_policy_set(Addon* callee, Promise* promise,
 // Returns IPCM_SUCCESS if a kernel plugin was successfully loaded/unloaded,
 // IPCM_FAILURE otherwise
 ipcm_res_t
-IPCManager_::plugin_load_kernel(Addon* callee, const std::string& plugin_name,
+IPCManager_::plugin_load_kernel(const std::string& plugin_name,
 				bool load)
 {
 	ostringstream ss;
@@ -1214,8 +1214,7 @@ IPCManager_::plugin_load(Addon* callee, Promise* promise,
 
 	try {
 		//First try to see if its a kernel module
-		if (plugin_load_kernel(callee, plugin_name,
-				       load) == IPCM_SUCCESS) {
+		if (plugin_load_kernel(plugin_name, load) == IPCM_SUCCESS) {
 			promise->ret = IPCM_SUCCESS;
 
 			return IPCM_SUCCESS;
--- a/rinad/src/ipcm/ipcm.cc
+++ b/rinad/src/ipcm/ipcm.cc
@@ -1174,14 +1174,10 @@ IPCManager_::plugin_load_kernel(const std::string& plugin_name,
 	} else if (pid == 0) {
 		// child
 		if (load) {
-			execl("modprobe", "modprobe", plugin_name.c_str(),
-			      NULL);
 			execlp("modprobe", "modprobe", plugin_name.c_str(),
 			       NULL);
 
 		} else {
-			execl("modprobe", "modprobe", "-r",
-			      plugin_name.c_str(), NULL);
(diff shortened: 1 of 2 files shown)


rinad/src/ipcm/ipcm.cc
```diff
@@ -1151,6 +1154,69 @@ IPCManager_::select_policy_set(Addon* callee, Promise* promise,
...
+		result = IPCM_FAILURE;
+
+	} else if (pid == 0) {
+		// child
+		if (load) {
+			execl("modprobe", "modprobe", plugin_name.c_str(), NULL);
+			execlp("modprobe", "modprobe", plugin_name.c_str(), NULL);
+
+		} else {
+			execl("rmmod", "rmmod", plugin_name.c_str(), NULL);
```
sandervrijders: Shouldn't this be modprobe -r?

vmaffione: Is it not the same? For me it's ok to switch "modprobe -r".

sandervrijders: Well, modprobe handles module dependencies as well. modprobe -r: If the modules it depends on are also unused, modprobe will try to remove them too.

I always see it as:
insmod / rmmod
modprobe / modprobe -r

vmaffione:
It's not an external API call, it's only internally used by the plugin_load() method. And it's intended to be used only in this way. plugin_load() methods hides the detail on whether the plugin is for kernel space code or userspace code.

vmaffione:
There is indeed. Open up the IPCM console and

>> plugin-load 4 red-pcc
>> plugin-unload 4 red-pcc

Check kernel logs to see that things really happen.

msune:
@vmaffione OK. I got confused by the Addon callee.

But you don't need the callee nor the promise, because the API is synchronous, so please remove them. 

We will have namespace problems with user-space kernel space naming of the policies. I suggest using something like:

`kernel:<policy_name>`
`user:<policy_name>`

We could even go beyond this and have a fully qualified name:

`kernel:<component>:<sub_component*>:..:<policy_name>`
`user:<component>:<sub_component*>:..:<policy_name>`

vmaffione:
@msune: We can choose to have that "kernel:" "user:" approach for namespaces, but this requires discussion (not in this PR).

vmaffione:
@msune : was that an ACK or not?

msune:
No it wasn't.  `040dc7a` partially addresses this, but:

* Callee is not needed by plugin_load_kernel()
* The parameterl `bool load` is strange; wouldn't it make more sense to have a plugin_load_kernel() and a plugin_unload_kernel()?

vmaffione:
Removed the callee argument as suggested, thanks.

Regarding the plugin_load_kernel() thing, it makes sense to have one in order to reuse code. It is the same technique already used in plugin_load() in IPCM and IPCP. So I would not change that in this PR.

edugrasa:
ACK, verified working

edugrasa:
@msune are you for an ACK or not convinced yet? :)

rinad/src/ipcm/ipcm.cc
```diff
@@ -1151,6 +1154,55 @@ IPCManager_::select_policy_set(Addon* callee, Promise* promise,
...
+		ss << "Kernel plugin (un)loading: fork() failed";
+		FLUSH_LOG(ERR, ss);
+		result = IPCM_FAILURE;
+
+	} else if (pid == 0) {
+		// child
+		if (load) {
+			execl("modprobe", "modprobe", plugin_name.c_str(),
+			      NULL);
+			execlp("modprobe", "modprobe", plugin_name.c_str(),
```
msune: Why 2 calls to exec? execlp should be enough I guess

(Commented also in the other PR, sorry for duplicating)

msune:
Once this comment about exec(s) is addressed, for me it would be an ACK.

Issue #677, opened 2015-07-10
IPCM: get_dif_template() shall not return a pointer


msune:
With the current `DIFTemplateManager::get_dif_template()` routine, the template manager exposes a pointer to the DIF template (internal pointer). Since the `DIFTemplateManager` exposes an API to remove templates, the pointer may not be valid once used. APIs like assignToDIF make use of these of pointers too.

The question that I have are:
- Is really necessary that the IPCM exposes an API to recover a DIF template object? Is it enough that IPCM exposes APIs like: 
  - `template_exists(const std::string& template_name)`
  - Use in all IPCM API calls `const std::string& template_name` as a handle for the template
  - [only if it is necessary]  get_template_info(): gets a copy of the template object

To fix the underlying problem, which can also be happening internally between `IPCManager` and `DIFTemplateManager`, options that I see:
- Do not allow to delete templates (in-memory)
- Expose a copy of the template always
- Implement ref count

msune:
@edugrasa thoughts?

edugrasa:
Fixed by #1086

PR #678, opened 2015-07-10
Add policies catalog and automatic plugin loading

vmaffione:
This patch series adds the following features:
- IPCM Catalog, implemented as a class contained in the main IPCManager_ class
- Automatic import of manifest files and loading of required userspace and kernelspace plugins, as specified by the DIF template
- IPCM Console command to show the catalog


vmaffione:
Maintainers: @edugrasa @msune

rinad/src/ipcm/ipcm.cc
```diff
@@ -1151,6 +1171,55 @@ IPCManager_::select_policy_set(Addon* callee, Promise* promise,
...
+		ss << "Kernel plugin (un)loading: fork() failed";
+		FLUSH_LOG(ERR, ss);
+		result = IPCM_FAILURE;
+
+	} else if (pid == 0) {
+		// child
+		if (load) {
+			execl("modprobe", "modprobe", plugin_name.c_str(),
+			      NULL);
+			execlp("modprobe", "modprobe", plugin_name.c_str(),
```
msune: Why 2 calls to exec? execlp should be enough I guess

rinad/src/ipcm/ipcm.cc
```diff
@@ -1151,6 +1171,55 @@ IPCManager_::select_policy_set(Addon* callee, Promise* promise,
 	return IPCM_PENDING;
 }
 
+// Returns IPCM_SUCCESS if a kernel plugin was successfully loaded/unloaded,
+// IPCM_FAILURE otherwise
+ipcm_res_t
+IPCManager_::plugin_load_kernel(Addon* callee, const std::string& plugin_name,
```
msune: callee is unnecessary here

rinad/src/ipcp/ipc-process.cc
```diff
@@ -419,7 +414,7 @@ void IPCProcessImpl::processSetPolicySetParamRequestEvent(
                 result = security_manager_->set_policy_set_param(remainder,
                                                                 event.name,
                                                                 event.value);
-        } else if (component == "enrollment") {
+        } else if (component == "enrollment-task") {
```
msune: Shouldn't this be addressed in another PR? Sounds like  abug

msune:
Question, and this is related to #677; is the catalog required to be protected with any rwlock? The same for the template manager.

rinad/src/ipcp/ipc-process.cc
vmaffione: It's just that we had not taken the decision yet ("enrollment" or "enrollment-task"). Now we have and so we use "enrollment-task" uniformly.

vmaffione:
@msune, yes the catalog needs to be protected by rwlock, since it's accessed during the assign_to_dif() in write mode - and more assign_to_dif() can execute in parallel, from scripting engine in write mode, and from console in read mode. I am adding the required locking.

vmaffione:
@msune, @edugrasa, can we merge this?

edugrasa:
making a quick test before acking

edugrasa:
In rinad/etc/ipcmanager.conf.in you should add the plugins path to local configuration, shoudn't you? something like:

  "localConfiguration" : {
    "installationPath" : "@prefix@/bin",
    "libraryPath" : "@libdir@",
    "logPath" : "@localstatedir@/log",
    "consolePort" : 32766,
    "pluginsPaths": ["@libdir@/rinad/ipcp"]
  },

vmaffione:
Yeah, right, I always forget that one, thanks. Doing that.

edugrasa:
I'm not sure about the logics of Catalogue.load_by_template.

Shouldn't it try to assemble a list of all the plugins that need to be loaded by the IPCP and then load them? Right now it checks policy-set by policy-set, gets the plugin the policy set is part of and calls load plugin. If you have 7 policy-sets belonging to the same plugin, the IPCM sends seven "plugin-load" netlink messages to the IPCP to load the same plugin (see log message below, it happens 6 times), when one would be enough.

9187(1436962791)#librina.nl-manager (DBG): NL msg TX. Fam: 25; Opcode: 58_PLUGIN_LOAD_REQ; Sport: 1; Dport: 9192; Seqnum: 3; Request; SIPCP: 0; DIPCP: 2
9192(1436962791)#librina.nl-manager (DBG): NL msg RX. Fam: 25; Opcode: 58_PLUGIN_LOAD_REQ; Sport: 1; Dport: 9192; Seqnum: 3; Request; SIPCP: 0; DIPCP: 2 
9192(1436962791)#librina.core (DBG): Added event of type 39_PLUGIN_LOAD and sequence number 3 to events queue
9192(1436962791)#librina.core (DBG): Waiting for message 9192
9192(1436962791)#rinad.event-loop (DBG): Got event of type 39_PLUGIN_LOAD and sequence number 3
9192(1436962791)#librina.application (INFO): Plugin 'default' already loaded
9192(1436962791)#librina.nl-manager (DBG): NL msg TX. Fam: 25; Opcode: 59_PLUGIN_LOAD_RESP; Sport: 9192; Dport: 1; Seqnum: 3; Response; SIPCP: 2; DIPCP: 0
9187(1436962791)#ipcm (INFO)[plugin_load]: Issued plugin-load to IPC process test1.IRATI:1::

Also, every time the ipcm gets a plugin-load response, there's this strange log:

9187(1436962791)#ipcm.policies (INFO)[ipc_process_plugin_load_response_handler]: plugin-load-op completed on IPC process test1.IRATI:1:: [success=1]

9187(1436962791)#ipcm (INFO): Plugin 'default' successfully loaded
FATAL: Module default not found.

vmaffione:
Agree, pushing the fix in short.

The log is because the IPCM first tries to see if the plugin is a kernel module. If it's not, it tries to see if it's a shared objeect (delegating to the IPCP). The log shows up when loading user plugins. We can decide to manage this in a different way (like prefixing the plugins with kernel: user:) but I would postpone this discussion in a future PR.

edugrasa:
Agree!

edugrasa:
Works well, it's an ACK for me!

vmaffione:
thanks, but please wait for a little more commit coming in short

vmaffione:
@msune, @edugrasa can we merge this?

rinad/src/ipcm/ipcm.cc
```diff
@@ -395,6 +400,26 @@ IPCManager_::assign_to_dif(Addon* callee, Promise* promise,
...
+			throw rina::AssignToDIFException();
+		}
+
+		ipcp = lookup_ipcp_by_id(ipcp_id, false);
+		if (ipcp->get_type() == rina::NORMAL_IPC_PROCESS) {
+			// Load all the plugins required from by template, but
+			// first release the ipcp lock, to avoid lock-ups.
+			ipcp->rwlock.unlock();
+
+			catalog.load_by_template(callee, ipcp_id, dif_template);
```
msune: I don't fully understand the rationale here. I think we should clearly separate the template manager from the catalog.

At first glance, I would say that the IPCM should be the only one who should "interact or bridge" between these two modules => do not embbed template logic stuff in the catalogue and viceversa.

In this case, what I think it should be done is that the IPCM recovers via the template (or whatever) the list of policies, and _then_ it actually loads them, making catalogue unaware of templates.

rinad/src/ipcm/ipcm.cc
```diff
@@ -1769,4 +1786,266 @@ void IPCManager_::io_loop(){
...
+	}
+
+	return suffix == s.substr(s.size() - suffix.size(), suffix.size());
+}
+
+#define MANIFEST_SUFFIX ".manifest"
+
+void Catalog::import()
+{
+	const rinad::RINAConfiguration& config = IPCManager->getConfig();
```
msune: Instead of a back-reference to the IPCManager and have to do call `getConfig()`, this could be part of the `Catalogue` constructor

rinad/src/ipcm/ipcm.cc
```diff
@@ -1769,4 +1786,266 @@ void IPCManager_::io_loop(){
...
+	ret = rina::plugin_get_info(plugin_name, plugin_path, new_policy_sets);
+	if (ret) {
+		LOG_WARN("Failed to load manifest file for plugin '%s'",
+			 plugin_name.c_str());
+		return;
+	}
+
+	if (plugins.count(plugin_name)) {
+		// Plugin already in the catalog
+		return;
```
msune: Shouldn't this send an exception?

msune: Probably this should be before the get_info call

rinad/src/ipcm/ipcm.cc
```diff
@@ -1769,4 +1786,266 @@ void IPCManager_::io_loop(){
...
+			// Nothing to do, we have already loaded this.
+			return 0;
+		}
+
+		plugin_name = cmap[psinfo.name].plugin->second.name;
+	}
+
+	// Perform the plugin loading - a blocking and possibly
+	// asynchronous operation - out of the catalog lock
+	if (IPCManager->plugin_load(addon, &promise, ipcp_id,
```
msune: I am very confused; it seems that the responsabilities of the IPCM and the Catalogue are not properly split. Calling the IPCM to load the plugin seems not the responsability of the Catalog.

My thoughts are:

* Catalogue: should solely have the "list" of available policy sets and policies (and the relation to policy set(s)), and its details (path, characteristics..).
* IPCM: in charge of consulting the catalogue, get the info, and load or unload the plugins.
* Who keeps track of the loaded or unloaded plugins? To me this is an IPCM class responsability.

My suggestion is to convert this method into a method that returns an exception or an `std::list<Plugin>`, so that the IPCM can call its internal methods to load.

msune:
As general comments:

* I would move the Catalog class into a dedicated catalog.h/cc. The ipcm.h/cc is big enough already.

rinad/src/ipcm/ipcm.cc
msune: -1 instead of exception is also ok

rinad/src/ipcm/ipcm.cc
vmaffione: Yes it's a good idea, I agree, but I think it can go in future pull requests. We need this PR in asap to start experimenting with non-default plugins.

rinad/src/ipcm/ipcm.cc
vmaffione: It's not a back-reference, it's just accessing the IPCM singleton.

rinad/src/ipcm/ipcm.cc
vmaffione: Regarding the get_info call, yes good catch --> changing it
Regarding the plugin_get_info, I would keep the "return -1" model, it's ok for our purposes.

rinad/src/ipcm/ipcm.cc
vmaffione: Again, I agree with you, but let's go step-by-step. These improvements you suggest can go in a future PR, so that in the meantime we can start experiment with non-default plugins.

vmaffione:
Regarding the separate compilation unit thing, I agree, and I'll try to do it now. (I didn't follow this way from the beginning because when trying to do that I ran into weird build problems, that I didn't have time to address, let's see what happens now).

vmaffione:
@msune, addressed your comments. Can you give a check?

edugrasa:
@vmaffione @lbergesio has just found a bug when IPCM creates two normal IPCPs and assigns them to a DIF. The "default" plugin is loaded for the first IPCP, but it is not loaded for the second one.

vmaffione:
@edugrasa Thanks for the feedback, going for a fix.

vmaffione:
@lbergesio @edugrasa, the problem you pointed out should be fixed now, can you give a test?

lbergesio:
problem is gone

rinad/src/ipcm/ipcm.cc
msune: This has to be addressed in this PR because this is new code (Catalog). NAK otherwise.

rinad/src/ipcm/ipcm.cc
msune: It is not a good idea to merge this PR as is, because it will be much more complicated to fix aftewards. This is new code!

NAK

vmaffione: As I have already said, this pull request is needed for experimentation, that we must start **immediately**.
This is why I want to merge it now - we can start experimentation in parallel with the improvements that you suggest.

It won't be complicated to fix afterwards - it's will just be the same. The other pull requests that are pending are implementation of new policy sets, that don't care at all about the IPCM internal implementation. It does not make sense to stall this for internal rearrangments that we have do anyway - since this catalog thing is basically part of second iteration SDK.

edugrasa: While I understand Marc's concerns, I agree with Vincenzo in that we need to move forward to go on with experimentation, even at the cost that the fix suggested by Marc could be more complicated afterwards.

msune: That's a one hour rework (or less).

But if you both think it is so urgent it cannot wait, just ACK it yourselves and merge it.

vmaffione: Self-ACK or partial-ACK it's against our rules, I won't do it.

All right, I guess who needs this (e.g. we need it in WP4/LFA) can merge it
in a local branch and work there for experimentation.

I'll try to consider your concerns when I've time - unless you want to fix
it yourself, in which case would be ok for me.

2015-07-17 18:02 GMT+02:00 Marc Suñé <notifications@github.com>:

> In rinad/src/ipcm/ipcm.cc
> <https://github.com/IRATI/stack/pull/678#discussion_r34903550>:
>
> > @@ -395,6 +400,26 @@ IPCManager_::assign_to_dif(Addon* callee, Promise* promise,
> >  	IPCPTransState* trans;
> >
> >  	try {
> > +		if (is_any_ipcp_assigned_to_dif(dif_name)) {
> > +			ss << "There is already an IPCP assigned to DIF "
> > +				<< dif_name.toString()
> > +				<< " in this system.";
> > +			FLUSH_LOG(ERR, ss);
> > +			throw rina::AssignToDIFException();
> > +		}
> > +
> > +		ipcp = lookup_ipcp_by_id(ipcp_id, false);
> > +		if (ipcp->get_type() == rina::NORMAL_IPC_PROCESS) {
> > +			// Load all the plugins required from by template, but
> > +			// first release the ipcp lock, to avoid lock-ups.
> > +			ipcp->rwlock.unlock();
> > +
> > +			catalog.load_by_template(callee, ipcp_id, dif_template);
>
> That's a one hour rework (or less).
>
> But if you both think it is so urgent it cannot wait, just ACK it
> yourselves and merge it.
>
> —
> <https://github.com/IRATI/stack/pull/678/files#r34903550>.
>



-- 
Vincenzo Maffione

msune: I can retire the NAK and edu is giving you the ACK, so no breaking of rules...

I simply cannot ack this pr as is without either a) getting convienced the proposed approach is the right one or b) address the issues.

Issue #675, opened 2015-07-09
RMT errs upon loading default config file


sandervrijders:
When one uses the config file that is installed together with the stack, the RMT complains that it is handed a bogus entry. Seems related to security. @olichtne @edugrasa 

```
[  389.272507] rina-kipcm(DBG): Found IPC Process with id 3
[  389.272510] rina-ipcp-utils(DBG): Copying name ffff8803269e2ea0 into ffff8800b7d93740
[  389.272512] rina-ipcp-utils(DBG): Name at ffff8800b7d93740 finalized successfully
[  389.272516] rina-ipcp-utils(DBG): Name ffff8803269e2ea0 copied successfully into ffff8800b7d93740
[  389.272518] rina-efcp(DBG): Succesfully set EFCP config to EFCP container
[  389.272520] rina-rmt(ERR): Bogus entry passed
[  389.272575] rina-rmt(DBG): Setting SDU protection policies to NULL
[  389.272578] rina-rmt(DBG): RMT PSs: default, lfa
[  389.294279] rina-kipcm(DBG): Assign to dif operation seems ok, gonna complete it
```

PR #679, opened 2015-07-13
Fixed two bugs in kernel space (non-present SDU protection config handling and destruction of normal over normal)

edugrasa:
* Fixed null pointer access when destroying a normal IPCP over a normal IPCP with a flow allocated
* Fix handling of not-present SDU protection configuration in kernel space (Fixes issue #675)

Maintainers: @lbergesio @miqueltarzan

### DIFF
--- a/linux/net/rina/ipcps/normal-ipcp.c
+++ b/linux/net/rina/ipcps/normal-ipcp.c
@@ -653,7 +653,7 @@ static int normal_assign_to_dif(struct ipcp_instance_data * data,
 
         sdup_config = dif_information->configuration->sdup_config;
         if (!sdup_config) {
-        	LOG_WARN("No SDU protection configuration specified");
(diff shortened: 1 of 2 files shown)


lbergesio:
@edugrasa: we still miss some infrastructure in the kernel for sdu protection, right? The way it is currently done in the n1_port_create and n1_port_write in rmt.c does not allow to plug different code without changing the "core". Anyway this is a parallel discussion...

this fix is ok, ACK

edugrasa:
Yes, sdk support is being addressed by Ondrej

lbergesio:
ack, fixed as discussed

miqueltarzan:
ACK

dstaesse:
Since it goes into the kernel, should this commit now be according to the kernel coding guidelines?
In that case, things need to be corrected.

PR #684, opened 2015-07-15
Bad locking in rmt.c when retrieving N-1 Port

lbergesio:
Simple locking fixing. @miqueltarzan is the mainteiner of the file, but I would ask @edugrasa and @olichtne to check I did not do anything wrong there.


edugrasa:
ack, verifired working

miqueltarzan:
ACK

PR #688, opened 2015-07-28
Completing LFA implementation in userspace

vmaffione:
This patch series contains the modification necessary to correctly compute LFA nodes for the routing table and correctly translate NextHop information into PortId information, so that the resulting PDU Forwarding table (that includes lists of alternate port-ids) is pushed to kernelspace to the PFF component.

### DIFF
--- a/librina/include/librina/ipc-process.h
+++ b/librina/include/librina/ipc-process.h
@@ -804,6 +804,14 @@ public:
         const std::string toString();
 };
 
+struct NHopAltList {
+	/** Next hop and its alternates */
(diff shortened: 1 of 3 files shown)


sandervrijders:
Testing it

sandervrijders:
ACK, works for the LFA case. Found a small bug in the LFA, proposing the fix in a different PR.

vmaffione:
Maintainers: @edugrasa @bernatgaston

bernatgaston:
ACK

edugrasa:
ACK

PR #689, opened 2015-07-28
plugins: lfa: Fixes bad index in for loop

sandervrijders:
This patch fixes a bug in lfa_add. The for loop always ended one entry
too soon, making some alternates disappear.

Maintainer: Myself.


vmaffione:
ACK

PR #691, opened 2015-07-28
Fix deployment of manifest files for kernel plugins

vmaffione:
With this patch, non-default plugin manifests are installed in the system in the same place compiled modules are installed in the system - under /lib/modules/`uname -r`.

In this way IPCM configuration file can specify a single entry in pluginsPaths to indicate where the manifest files for kernel plugins can be found.

Maintainers:
  @lbergesio @msune


msune:
ACK for maintainers

edugrasa:
ACK

PR #696, opened 2015-07-29
Fix and cleanup of build system

msune:
Contains:

* Fix CPPFLAGS != CXXFLAGS
* Remove unnecessary checkings and old .m4 files which produce warnings in automake 1.14 and above.
* Remove -Wextra from default and remove unnecessary (void)
* Remove attic folders from rinad and rina-tools

Maintainers:

@vmaffione @edugrasa @bernatgaston


msune:
I recommend checking the patch *before* last commit (remove attic), so until 72c0ac8b29e0d2b51ed8287e5c2184439e496e35

vmaffione:
Why should we check not considering the last commit? Can you elaborate a bit?

msune:
No, sorry. I meant the last commit adds a lot of removes to the diff, so it hides the real part of changes in the first 3 commits.

I was suggesting to check the 3 commits first, then check the last one (remove atics).

Sorry for the bad wording

vmaffione:
Ah ok, got it. For me it's an ACK, compilation works ok, and tests are ok.

msune:
Which spelling issue?

Please test the compilation, both with JAVA bindings and without. Tests are not OK due to issues in pristine-1.3. Works for me aside of make check.

msune:
Found your comment. Fixing it.

msune:
Fixed and fixed 2nd commit message that had typos. Rebased.

vmaffione:
I've compiled userspace parts using ./install-user-from-scratch, that enables librina java bindings. So it is ACK for me.

edugrasa:
ACK

msune:
I consider 2 ACKs (+ mine) sufficient so merging

PR #697, opened 2015-07-29
rinad: ipcm: Sets a better default configuration file path 

### LINKED ISSUE #672: ipcm: Load a default config file
Upon start, the ipcm requires a config file. Currently one always has to specify it:

```
   -c <string>,  --config <string>
     (required)  Configuration file to load
 ```

It would be handy for testing purposes to have it point to a default location, such as /etc/ipcmanager.conf, or $SANDBOXED_ENVIRONMENT/etc/ipcmanager.conf.

### PR DESCRIPTION
**Pull request by sandervrijders**
This patch sets a better default path for the ipcmanager.conf
configuration file. It uses the prefix that was passed on installation
as the base path followed by /etc/ipcmanager.conf

Fixes #672 

Maintainer: @msune

### CODE REVIEW

File: rinad/src/ipcm/Makefile.am [hunk 1]
```diff
@@ -26,7 +26,8 @@ ipcm_CPPFLAGS =					\
 	-I$(srcdir)/..				\
 	-I$(srcdir)/../common			\
 	-DPLUGINSDIR=\"$(pkglibdir)/ipcm\"	\
-	-DIPCPPLUGINSDIR=\"$(pkglibdir)/ipcp\"
+	-DIPCPPLUGINSDIR=\"$(pkglibdir)/ipcp\"  \
+	-DPREFIX=\"$(prefix)\"
```
msune: No need for this in theory; PREFIX should be already in CPPFLAGS.

msune:
ACK

PR #698, opened 2015-07-29
rinad: ipcm: added graceful exit upon SIGQUIT, SIGTERM, SIGHUP

dstaesse:
ipcm manager now exits cleanly upon killing through the kill command and accidental ssh disconnects.

easiest way to test this PR, start your ipcm and then try the following commands
```
sudo kill <pid>
sudo kill -SIGINT <pid>
sudo kill -SIGHUP <pid>
sudo kill -SIGTERM <pid>
sudo kill -SIGQUIT <pid>
```
and restart the ipcm after each.

Maintainer: @msune

### DIFF
--- a/rinad/src/ipcm/main.cc
+++ b/rinad/src/ipcm/main.cc
@@ -57,6 +57,9 @@ void handler(int signum)
 			dump_backtrace();
 			exit(EXIT_FAILURE);
 		case SIGINT:
+		case SIGQUIT:
+		case SIGTERM:
+		case SIGHUP:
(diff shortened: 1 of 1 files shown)


msune:
Weird indentation of lines:

```
+        LOG_DBG("SIGHUP handler installed successfully");
```

But I see that this comes from the original traces. So ACK

dstaesse:
@msune: yeah that's what I mentioned on skype. I don't see this weird indentation in emacs, it's a hard/soft tab mismatch which shows int the github interface.

PR #703, opened 2015-07-30
Pristine 1.3 jain dtcp: fixes to the policy set

edugrasa:
a) Removed the code to check for duplicated PDUs (doesn't belong to this policy and was causing memory corruption)

b) Avoid starvation of flows by forcing the minimum window to 1 (not to 0)

c) Use real numbers (in fixed point, not floats) to calculate windows, then round to integer (as suggested by Jain)

Maintainers: @lbergesio


lbergesio:
NACK

compiling the module fails to me:
WARNING: "__divdi3" [/home/irati/plugins/cong_avoidance/cas-plugin.ko] undefined!

Moreover, some extra cleaning is required. I will create a new PR with this plus the required changes.

sandervrijders:
@lbergesio Close this one?

PR #705, opened 2015-07-31
Pristine 1.3 xmit block: better integration with dev-queue-xmit and net devices

edugrasa:
Changes to RMT:
 * if writing to the N-1 flow returns -EAGAIN, requeue PDU, mark N-1 port as DISABLED.

Changes to shim DIF:
 * The only way to get a clue on congestion at the netdevice and get notifications of when the device is ready to transmit again without modifying linux kernel source files is to attach a Qdisc to the physical device driver where packets of the shim DIF will go through (i.e. eth0, eth1, etc..).

* This Qdisc has a queue that is populated and drained by the traffic control layer, based on feedback from the device driver. When the queue of the Qdisc is full, dev_queue_xmit will return error and the shim will return -EAGAIN on the write operation (RMT will mark N-1 flow as DISABLED as explained in the changes to the RMT). When the queue starts to decrease (due to data being transmitted by the device) and reaches a certain threshold, the qdisc will tell the shim ethernet to re-enable all the flows that were going through that device. When this happens affected RMTs will schedule the tasklets to empty their queues, writing to N-1 flows again until RMT queues are empty or write returns -EAGAIN.

Maintainers: @sandervrijders, @miqueltarzan. It would be nice if @vmaffione could also have a look.


linux/net/rina/ipcps/shim-eth-vlan.c
```diff
@@ -1328,18 +1589,38 @@ static int eth_vlan_update_dif_config(struct ipcp_instance_data * data,
...
-                }
+                if (!strcmp(entry->name, "interface-name")) {
+                	ASSERT(entry->value);
+
+                	info->interface_name = rkstrdup(entry->value);
+                	if (!info->interface_name) {
+                		LOG_ERR("Cannot copy interface name");
+                		return -1;
+                	}
+                } else if (!strcmp(entry->name, "qdisc-max-size")) {
```
sandervrijders: Is it needed to be able to specify this in the DIF template?

linux/net/rina/serdes.c
```diff
@@ -791,8 +791,6 @@ static struct pdu_ser * pdu_serialize_gfp(gfp_t                       flags,
             ASSERT(dup_chksum_is_ok(tmp));
         }
 
-        pdu_destroy(pdu);
-
```
sandervrijders: Why is this removed? pdu_serialize takes ownership of the pdu, converts it into a pdu_ser and destroys the pdu.

linux/net/rina/ipcps/shim-eth-vlan.c
edugrasa: It's optional If you don't do it it uses the defaults, otherwise what is in the template. I've been using this code for the last month, but still not 100% sure on the influence of these values on performance. Therefore it's better that they are configurable for now, to facilitate experimentation with different values.

linux/net/rina/serdes.c
edugrasa: Yeap, and this is wrong. Writing to the N-1 port can return -EAGAIN (try later, congestion now), and then the PDU has to be requeued. If the PDU has already been destroyed by pdu_serialize, you can't requeue it.

linux/net/rina/ipcps/shim-eth-vlan.c
sandervrijders: Ok, that's fine.

linux/net/rina/serdes.c
sandervrijders: Aha, I see. So ownership is no longer passed. That's ok.

sandervrijders:
Looks OK at first sight, will take a better look now by porting it into my branch. And testing it.

Minor point: At the devs meeting we agreed on the need for the commits to also have an "in-body" description. http://chris.beams.io/posts/git-commit/

edugrasa:
Right Sander, forgot about it. I'll start using it in the next commits :)

sandervrijders:
Ok, thanks.

linux/net/rina/rmt.c
```diff
@@ -881,35 +915,52 @@ static void send_worker(unsigned long o)
...
-                        reschedule++;
-
                 rcu_read_lock();
                 if (ps && ps->rmt_next_scheduled_policy_tx) {
+                	pdus_sent = 0;
+                	ret = 0;
                         do {
-                                pdu = ps->rmt_next_scheduled_policy_tx(ps, n1_port);
-                                if (pdu) {
+                        	pdu = ps->rmt_next_scheduled_policy_tx(ps, n1_port);
```
sandervrijders: This line is > 80  and several others as well in this function (line 929, 931, 935, 936, 949). Please fix it.

linux/net/rina/ipcps/shim-eth-vlan.c
```diff
@@ -713,6 +744,182 @@ static int eth_vlan_application_unregister(struct ipcp_instance_data * data,
...
+	unsigned long               flags;
+
+	if (!data) {
+		return;
+	}
+
+	spin_lock_irqsave(&data->lock, flags);
+	list_for_each_entry(flow, &data->flows, list) {
+		if (flow->user_ipcp && flow->user_ipcp->ops) {
+			flow->user_ipcp->ops->enable_write(flow->user_ipcp->data,
```
sandervrijders: This line is too long.

linux/net/rina/ipcps/shim-eth-vlan.c
```diff
@@ -1227,8 +1437,38 @@ static int eth_vlan_assign_to_dif(struct ipcp_instance_data * data,
                                 data->dif_name = NULL;
                                 return -1;
                         }
+                } else if (!strcmp(entry->name, "qdisc-max-size")) {
+                	ASSERT(entry->value);
+
+                	result = kstrtouint(entry->value, 10, &temp);
+                	if (result) {
+                		LOG_ERR("Cannot conver qdisc-max-size to uint");
```
sandervrijders: This line is too long.

sandervrijders:
General note: There seems to be a mixture of hard and soft tabs in the new code (e.g. on the same line). Please use hard tabs only.

linux/net/rina/ipcps/shim-eth-vlan.c
msune: Be careful when trying to fix this. The string hall NOT be shortened. 

String traces cannot be split, to make them grepable. Check this exception in kernel's ciding style.

dstaesse: conver -> convert, same in L.1605, 1614

edugrasa:
@sandervrijders fixed the hard vs. soft tabs in the new code, as you say would be nice to fix in the new code

dstaesse:
@edugrasa there's still a "conver" in L.1608

edugrasa:
@dstaesse fixed

miqueltarzan:
ACK

vmaffione:
Usual tests seem to work

linux/net/rina/ipcps/shim-eth-vlan.c
```diff
@@ -713,6 +744,176 @@ static int eth_vlan_application_unregister(struct ipcp_instance_data * data,
...
+
+        list_for_each_entry(pos, &(eth_vlan_data.instances), list) {
+                if (pos->phy_dev == dev)
+                	eth_vlan_enable_all_port_ids(pos);
+        }
+}
+
+/*
+ * Private data for a rina_shim_eth scheduler containing:
+ * (only supports a single flow for the moment)
```
vmaffione: What do you mean by "only supports a single flow"?

vmaffione:
Patch looks ok but I think we need to test that actually does the job. Can we prove (with experimental tests) that with this patch the flow control between shim-eth and NIC driver actually works? In other words, we need a situations where we oversaturate a NIC, so that queuing kicks in and we can see that the system work well.

edugrasa:
Yes, Ive tested that. With XEN you can rate limit the interfaces of the VMs
easily when using paravirtualized drivers. More tests and experimentation
are needed in the Vwall to try different environments, agree, but the
approach works well. The qdisc code can be fine tuned (when to call enable
all ports, basically).

vmaffione:
But what do you mean by only "it only supports a single flow for now"?

lbergesio:
@vmaffione if I recall well, it is just that the shim-eth-vlan only supports one application on top requesting 1 flow at a time so that the "enable N+1 ports" will actually be 1 N+1 port.

If this is clear this PR could be merged right?

vmaffione:
Not exactly. The shim-eth vlan supports only registration (so only one application), but multiple flows from that applications to applications on the same LAN segment. In other words, shim-eth-vlan it's not a point-to-point DIF (like the shim-hv). If this patch supports "just one flows" (still have to understand what this really means), it means that a simple scenario with three nodes on the same LAN segment will not work - e.g. it would not be possible to allocate a flow between A and B, one between B and C and one between A and C.

vmaffione:
In every case, the code seems to enable all the possible N+1 port-ids, when needed, and not just one. So perhaps is that comment outdated?

lbergesio:
You are right. I do not get the comment now. Since the shim-eth-eth will have only have 1 qdisc although it supports several N+1 flows (from the same app) you have to disable and enable all the ports so all the N+1 ports. And the code does that... maybe it is outdated or we are missing something. I will try to clarify it today

lbergesio:
@vmaffione As we said, that was an old comment, PR should be fine now!

vmaffione:
Yeah, sorry, but >1 is more expressive, imho :)

However, that was perfectly clear to me, I was asking a more specific thing: Why is the check >=2 and not >=1? I guess because when you call the restore_qdisc() you still have to remove the current shim IPCP from the list of shim eth IPCPs. Is that right?

lbergesio:
Not sure if I got you, but... you always remove the current shim ipcp from the list. The point is to restore the default qdisc to the phy_dev or not. If you use >=1 in the case of 1 you would be not restoring the qdisc, and you should do it because it is the only and last shim ipcp in that phy dev which will be destroyed. If the question is why ">=2" instead of ">1" there is no reason....

vmaffione:
I think we are not understanding each other.
When you delete a shim-eth-vlan IPCP, obviously, you remove the IPCP from the list of shim-eth-vlan IPCPs. Question was simply: can you confirm the removal of the IPCP from the list happens right after the restore_qdisc() function is called? If you confirm that, then the check >=2 is correct, otherwise it's mistaken.

lbergesio:
eth_vlan_destroy is the same one, it justs calls eth_vlan_restore_qdisc to, if possible, restore the qdisc. eth_vlan_restore_qdisc has nothing to do to deleting ipcps. So, I can confirm your question and that the ">=2" is correct.

edugrasa:
Hi Vincenzo, yes

vmaffione:
Ok, I think it's a good job :)

PR #710, opened 2015-08-07
Pristine 1.3 jain dtcp ps

lbergesio:
This PR fixes and extends #703 which was originally nacked by me.

It implements the Jain binary congestion avoidance scheme as a pair of PSs for DTCP and RMT.
It has been tested and used in the first experiment scenario of D6.2.

Mainteiners: @miqueltarzan and myself.

### DIFF
--- a/plugins/cong_avoidance/dtcp-ps-cas.c
+++ b/plugins/cong_avoidance/dtcp-ps-cas.c
@@ -35,23 +35,15 @@
 #define W_INC_A_P_DEFAULT     1
 #define W_DEC_B_NUM_P_DEFAULT 7
 #define W_DEC_B_DEN_P_DEFAULT 3
-#define BITS_PER_BYTE 8
-#define BITS_PER_INT (sizeof(int) * BITS_PER_BYTE)
-#define VECTOR_SIZE(X) ((((X) / BITS_PER_INT) + 1) * sizeof(int))
-#define BIT_INDEX(X) ((X) / BITS_PER_INT)
-#define BIT_NUMBER(X) ((X) % BITS_PER_INT)
 
 struct cas_dtcp_ps_data {
-        bool         first_run;
         seq_num_t    wc;
         seq_num_t    wp;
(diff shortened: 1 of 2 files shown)


miqueltarzan:
ACK for RMT

msune:
Not meaning to ammend the commit messages for this PR. But please have a look on the mail circulated for having better (more understandable and traceable) commit messages:

http://chris.beams.io/posts/git-commit/

The first line of a commit message is the subject, and should be <= 50 characters, and then have a body with the real explanation. Vi and other editors, when set to be the main git editor, they show when you go out of bounds for subject and body of the commit message.

lbergesio:
You are right Marc, several commits here should not be there, if you think it is needed I can squash a lot of things that were done by experimentation and I introduced in the PR when doing merge. 

For the first line with <= is sometimes can of difficult if you also have to identify the component, i.e. "kernel: cas-plugin: rmt: " is around 25 chars already. But agree with your comment.

lbergesio:
@msune or @sandervrijders since the rest of the people with merge rights seem to be in holidays, could any of you please merge this PR? changes in RMT are minimum and they are acked, the rest is self-contained in the jain's PS which I have been testing and works.

sandervrijders:
@lbergesio okay, but I'd prefer it if you squashed some of these commits and issued another PR.

lbergesio:
Ok, I thought it was not necessary since I proposed it, will do

msune:
As I was saying, it would be better but I don't think is imperative for this PR.

Btw, what you DON'T have to do is to create a new PR. Remember, pull request are against branches.. you can rebase (squash) in that branch (push force) and the PR will be updated.

sandervrijders:
Right, rebase in that branch should also work instead of squash merging in a different branch. 

12 times simply the commit message "logs", sorry, that's a bit too much for my taste.

lbergesio:
This should be ok now. #703 to be closed after merge too.

PR #712, opened 2015-08-12
Added AS_IF to check that actually the binaries are found

### LINKED ISSUE #709: librina installation fails because AC_PATH_PROG  of PROTOC is not being checked
Hi,

I am trying to install the sdk and I fell into this error:
```
rm -f CDAP.tmp
touch CDAP.tmp
/bin/false -I. -I/home/bird/irati/librina/src	\
	--cpp_out=.		\
	/home/bird/irati/librina/src/CDAP.proto
Makefile:1695: recipe for target 'CDAP.stamp' failed
```
This is caused because during configuration the check of the PROTOC path is failing:
```
checking for management_sdu_write syscall availability... yes
checking for protoc... /bin/false
checking for LIBPROTOBUF... yes
```
The root problem seems to be that the assigned path is not being checked:
https://github.com/IRATI/stack/blob/master/librina/configure.ac#L240

Other paths that are not being checked are:
https://github.com/IRATI/stack/blob/master/rinad/configure.ac#L28-L31
https://github.com/IRATI/stack/blob/master/rina-tools/configure.ac#L31-L34
https://github.com/IRATI/stack/blob/master/librina/configure.ac#L32-L35

vicalro:
Added AS_IF checks and error messages for configuration of librina, rinad, rina-tools.

fixes #709

### DIFF
--- a/librina/configure.ac
+++ b/librina/configure.ac
@@ -29,10 +29,22 @@ AC_PROG_GREP
 AC_PROG_MKDIR_P
 AC_PROG_LN_S
 
-AC_PATH_PROG([FMT],  [fmt],  [/bin/false])
-AC_PATH_PROG([SORT], [sort], [/bin/false])
-AC_PATH_PROG([UNIQ], [uniq], [/bin/false])
-AC_PATH_PROG([GIT],  [git],  [/bin/false])
+AC_PATH_PROG([FMT],  [fmt],  [no])
+AS_IF([test "x$FMT" = "xno"],[
+	AC_MSG_ERROR([Your system lacks of fmt support])
+])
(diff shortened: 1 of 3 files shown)


sandervrijders:
@vicalro Please specify the maintainers. (check_maintainers.py script in the root) Giving it a try.

sandervrijders:
ACK

Without git:
```
checking for git... no
configure: error: Your system lacks of git support
```

With git:
```
checking for git... /usr/bin/git
```

vicalro:
Need the ack from the maintainer of build system of librina, rinad & rina-tools: @msune

msune:
ACK

PR #715, opened 2015-08-18
pristine-1.3: kernel: dtcp: check ps->rtt_estimator is not NULL

lbergesio:
Nothing prevents a PS to not define the rtt_estimator op and set it
to NULL. If this happens, the DTCP code executes the op without checking
if the pointer is not NULL and the system crashed.
The proper solution would be to check for mandatory ops in the PS, but in
the meanwhile this solves the problem.

I am the maintainer, but changes are trivial so anyone with rights can check and merge.


sandervrijders:
ACK.

Yeah, that's probably part of re-evaluating the SDK. I guess some operations are always needed.

Issue #716, opened 2015-08-20
pristine-1.3: soft lock up while enabling/disabling N-1 port


lbergesio:
I just got the trace error attached here. There was a case where the shim-eth interrupt can lockup the code getting the lock of the n1_ports/n1_port structs.

[  488.212057] BUG: soft lockup - CPU#0 stuck for 22s! [rina-tgen:4110]
[  488.212057] Modules linked in: red_plugin(O) normal_ipcp rina_default_plugin shim_eth_vlan rinarp arp826 rina_personality_default 8021q garp stp llc nfsd auth_rpcgss oid_registry nfs_acl nfs lockd fscache sunrpc psmouse i2c_piix4 pcspkr serio_raw hid_generic usbmouse usbhid hid floppy
[  488.212057] CPU: 0 PID: 4110 Comm: rina-tgen Tainted: G           O  3.15.0 #19
[  488.212057] Hardware name: Xen HVM domU, BIOS 4.1.4 08/17/2014
[  488.212057] task: ffff880029e09840 ti: ffff88002ba3c000 task.ti: ffff88002ba3c000
[  488.212057] RIP: 0010:[<ffffffff810013a8>]  [<ffffffff810013a8>] xen_hypercall_sched_op+0x8/0x20
[  488.212057] RSP: 0018:ffff88003fc03a08  EFLAGS: 00200206
[  488.212057] RAX: 0000000000000000 RBX: 0000000000000000 RCX: ffff88003dc004e0
[  488.212057] RDX: 0000000000000000 RSI: ffff88003fc03a18 RDI: 0000000000000003
[  488.212057] RBP: ffff88003fc03a38 R08: 0000000000000000 R09: ffff88003dc00490
[  488.212057] R10: 0000000000000000 R11: 0000000000000000 R12: ffff88003fc03978
[  488.212057] R13: ffffffff8171b8ad R14: ffff88003fc03a38 R15: 0000000000000000
[  488.212057] FS:  0000000000000000(0000) GS:ffff88003fc00000(0063) knlGS:00000000f6e1c700
[  488.212057] CS:  0010 DS: 002b ES: 002b CR0: 0000000080050033
[  488.212057] CR2: 000000000906f000 CR3: 00000000302d8000 CR4: 00000000001006f0
[  488.212057] Stack:
[  488.212057]  ffffffff814300ee 00000009810c03e7 ffff88003fc03a14 ffffffff00000001
[...]

PR #718, opened 2015-08-21
pristine-1.3: fix softlockup in enable/disable port 

lbergesio:
This solves #716 
the interrupt coming from the NIC when receiving a frame
can take the lock in the n1_ports/n1_port structs in the rmt
when trying to enable/disable it. The same operation can be concurrently
originated by a write syscall from user-space. irqsave added to the lock ops.

Maintainer @miqueltarzan


miqueltarzan:
ACK

PR #719, opened 2015-08-25
Fix catalog for automatic loading of policy-sets

vmaffione:
The processing of DTP and DTCP policy set configurations was
missing.

Maintainers:
@lbergesio 
@msune

@kewinrausch, this will solve your problem


kewinrausch:
Vincenzo PR acked. Now the policy(in my case was DTP one) is successfully loaded by the policy catalog.

vmaffione:
@lbergesio , could you please ACK this? (I'ts just a line on your side)

I'll merge this in short since it's an easy bugfix

edugrasa:
ack on behalf of @lbergesio, since it is a simple one and he is on holidays this week

PR #727, opened 2015-09-03
kernel: shim-eth-vlan: fixed bug in qdisc when device had multiple tx queues (only one was replaced)

edugrasa:
Sander uncovered a bug in the qdisc used by the shim-eth-vlan: when the device had multiple tx queues only the first one was replaced (now all of them are). 

Also updated the default qdisc qsize to 500, since it seems a safe limit after testing in the VWall. The function of this qdisc is to cause the ports that use the physical device the qdisc is attached to to be blocked when the queue is full, and to unblock when the queue is almost empty (to avoid losing PDUs when the net device approaches saturation). 
Since the dev_queue_xmit operation cannot be called holding any locks, the sequence RMT write->call dev_queue_xmit, queue is full, shim returns -EAGAIN RMT blocks port can complete later than what it takes for the qdisc's queue to be drained and reach the "enable threshold", therefore the port may be enabled before it is disabled (it is a timing issue, race condition). The bigger is the difference between the qdisc q size and the enable threshold, the less likely this is to happen since it takes longer for the queue to be emptied (500 and 10 seem to be a safe value after experiments). Adding a timer that would unlock the port as a safety measure could be a plus, or other improvements/solutions for this issue can be considered after the review (with the limitation that we cannot change any of the Linux source code). The mechanism is ok for our current needs.

### DIFF
--- a/linux/net/rina/ipcps/shim-eth-vlan.c
+++ b/linux/net/rina/ipcps/shim-eth-vlan.c
@@ -862,6 +862,8 @@ int eth_vlan_update_qdisc(struct net_device *    dev,
 		       shim_eth_qdisc_ops.id) == 0)
 		return 0;
 
+	old_qdisc = dev->qdisc;
+
 	for (q_index = 0; q_index < dev->num_tx_queues; q_index++) {
 		sch = qdisc_create_dflt(netdev_get_tx_queue(dev, q_index),
 					&shim_eth_qdisc_ops, 0);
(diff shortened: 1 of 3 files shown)


sandervrijders:
Please split this into 2 PRs: the bugfix for the shim-eth-vlan and the changes in RMT (which should build upon the PR that refactored RMT to avoid merge horrors).

Also, what happens if I set q_enable_thres to zero in the shim-eth-vlan?

edugrasa:
If you set q_enable_thres to zero it will call i) enable_port too many times, and ii) you will be waiting too much to enable the port again.

What if we wait to merge this PR until the RMT one is merged first?

sandervrijders:
Yes, then you will get merge conflicts that have to be resolved. Either way works for me, but then the RMT PR should be merged soon (since we need this bugfix on the vwall).

PR #738, opened 2015-09-16
kernel: shim-eth-vlan: separated qdisc handling code from the shim-eth-vlan

### LINKED ISSUE #723: shim-eth: investigate if qdiscs can be moved to separate source file shim-utils-eth.[c/h]
Since the qdiscs can be applied to any shim DIF over ethernet, it may make sense to move them outside the source file. I think that was the rationale for prefixing public functions with shim_eth_?

edugrasa:
In order to allow qdisc utils to be re-used by other shim-eths, they have been
moved out to a separate file, shim-eth-utils.c. A header file with declarations
of the required operations has also been added shim-eth.h, and shim-eth-vlan.c
had to be renamed to shim-eth-vlan-main.c. This is becasue the shim-eth-vlan
kernel module now has multiple compilation units (shim-eth-vlan.o and shim-eth-
utils.o), and the name of the kernel module cannot be the same as the name
of one of its compilation units, therefore now the modules is still called
shim-eth-vlan and has shim-eth-vlan-main.o and shim-eth-utils.o as compilation
units (thanks to Vincenzo for the explanation).

When we add another shim-eth we'll have to slightly modify shim-eth-utils to
accept a function pointer, since each shim-eth-x will have a different
implementation of the enable-write-all function.

Fixes #723 
Maintainers: @sandervrijders


linux/net/rina/ipcps/Makefile
```diff
@@ -10,5 +10,7 @@ ccflags-y += -Wtype-limits
 
 obj-$(CONFIG_RINA_NORMAL_IPCP)    += normal-ipcp.o
 obj-$(CONFIG_RINA_SHIM_ETH_VLAN)  += shim-eth-vlan.o
+shim-eth-vlan-y :=				\
```
vmaffione: This should probably be

shim-eth-vlan-$(CONFIG_RINA_SHIM_ETH_VLAN) :=

otherwise I think you get an error when not compiling in shim-eth-vlan

edugrasa: No, I've checked https://www.kernel.org/doc/Documentation/kbuild/makefiles.txt, sections 3.2 and 3.3. If the shim-eth-vlan is not enabled for compilation the makefile will be

obj-n += shim-eth-vlan.o 

and the module will be ignored. The second line is just in case the module has multiple files, and it always needs to be <module name>-y.

I tried with the replacement and produced an error when compiling (that's why I checked the syntax).

dstaesse:
@edugrasa: very minor comments: 
Keep commit messages to the 50 char limit.
What about naming it shim-eth-vlan-core.c instead of shim-eth-vlan-main.c (it has no "main()")?

vmaffione:
@dstaesse 
I think it's ok like that. It's common to have in kernel code xxxx-main.c, this is not related to the main() function. (main --> main module). See for instance e1000 directory in drivers/net/ethernet.

sandervrijders:
@vmaffione Yeah, but we also have arp826-core.c in rinarp. Better to be consistent and change one of the 2.

edugrasa:
So, main or core?

vmaffione:
I think it's the same. I see here preference for core, so let's go that direction

sandervrijders:
ACK

PR #739, opened 2015-09-16
Track selected policy sets

vmaffione:
This PR introduces the following changes:
   - Add support to keep track of which policy-set is currently being selected for a given component of an IPCP.
   - Improved the show-catalog IPCM console command to show the new information
   - Improved the show-catalog IPCM console commend to show only information related to a single component (e.g. show-catalog pff).
   - Reworking of the catalog implementation, to use pointers instead of iterators.
   - Add update-catalog command, to rescan the pluginsPaths directory and update the catalog.
   - Some minor fixes to configuration files and implementation of a simple toString() method in librina

You can test by using select-policy-set command in the IPCM console and examining how the show-catalog output changes accordingly.

MAINTAINERS:
@edugrasa 
@msune

### DIFF
--- a/librina/include/librina/application.h
+++ b/librina/include/librina/application.h
@@ -165,7 +165,7 @@ struct PsInfo {
 	PsInfo(const std::string& n, const std::string& c,
 	       const std::string& v) : name(n), app_entity(c), version(v) { }
 
-	std::string toString() const { return name + std::string("/") + app_entity; }
+	std::string toString() const { return app_entity + std::string("/") + name; }
 };
 
 struct PsFactory {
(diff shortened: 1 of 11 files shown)


edugrasa:
Nice functionalities, ACK. It would be nice if you could add a little "usage" string in the show-catalog console command in the case that 2 arguments or more are provided to the command, like
USAGE: "show-catalog" or "show-catalog <component-name>"

vmaffione:
@edugrasa , Addressed your comment.

Remember that console has "help" command, so that in any case you can
````
help show-catalog
````
to get the usage.

PR #741, opened 2015-09-17
Userspace plugins: replace init() function with get_factories()

vmaffione:
With this patch, the plugin init() function is replaced by the get_factories() function. In this way, the IPCP acts as a master and asks the plugin for its policy sets. This simplifies the implementation and remove thread-safety-related problems.

This PR depends on #739 (includes it).

MAINTAINERS:
@edugrasa


edugrasa:
ACK, confirmed working.

PR #750, opened 2015-09-21
kernel: rmt: fixed race condition enabling/disabling ports

edugrasa:
When the shim-eth-vlan returned -EAGAIN, the qdisc code might call enable
on the N-1 port again before the rmt had time to disable it, resulting in
the port being blocked forever. This race condition has been fixed by this
commit, introducing an extra state in the N-1 port as seen by the RMT.

Maintainers: @miqueltarzan 
@sandervrijders could you give it a go in the Vwall?


linux/net/rina/rmt.h
```diff
@@ -59,6 +59,7 @@ struct rmt;
 enum flow_state {
 	N1_PORT_STATE_ENABLED,
 	N1_PORT_STATE_DISABLED,
+	N1_PORT_STATE_DO_NOT_DISABLE,
```
dstaesse: NACK. We should not add a state just to solve an implementation issue.

edugrasa: What do you mean? 

All the states are an implementation issue. It is the only way to solve the race condition, do you have a better alternative?

edugrasa: Moreover, imho the first person that should ACK/NACK the code based on design aspects is the maintainer of the component, since he/she is the person that knows more about the design of that particular component and about the expected usage of its data structures and functions. 

Otherwise we are defeating one of the points of having maintainers, in my opinion.

dstaesse: All I mean is that it looks to me that this PR is abusing the flow_state enum to share a variable between some functions. The flow_state is not (or at least should not be) implementation specific, it should be defined independently of the implementation and reflect the state machine for the flow/port. N1_PORT_STATE_DO_NOT_DISABLE does not seems like something that's fundamental for the FSM of a flow. For the solution to this issue, you can, for instance, define a variable outside of the flow_state enum, that's all.

msune:
My 2 cents,

First I agree with Eduard that we have to be careful before giving NACKs (and how we do peer reviewing in general).

IMHO when sending a NACK one should _either_ have a **very** clear idea that the changeset is completely wrong, name precisely why and suggest an alternative solution _or_ ask for more information on why it is done this way to the author/s. Otherwise we run the risk to upset contributors and just produce additional delays. 

Now about the PR, @edugrasa if (and only if) you think I can be helpful in here, I would need to understand which exact part of the code can trigger an enable (from the qdisc code) and interfere with the disable (IOWs which are the general events which can trigger an enable/disable of the port).

dstaesse:
@msune, @edugrasa: didn't mean offence, it's just raising an issue on my account. If all maintainers ACK, and merges this PR in, I wouldn't mind. But I see no reason to wait for people to flag things which they think can be improved. Maybe I should not have stated the "NACK." but just flagged the issue, I'll keep that in mind next time.

miqueltarzan:
@dstaesse no, flow state for RMT N-1flows has nothing to do with the FSM of flows; instead, the state here captures the local, current situation of that flow in that specific system, like the inability to go on writing since the layer below is overwhelmed. Defining a variable outside the enum would be a very unorthodox way of creating a new state for the N-1flow outside the state variable (the enum).

I have to test it but the code looks ok..

linux/net/rina/rmt.h
edugrasa: A variable that represents one of the states of an N-1 port as seen by the RMT (this is all this flow state enum is).. well the best place to put it it's the flow state precisely in my opinion. It is another state of the N-1 port, it means that the N-1 IPCP already re-enabled the port and therefore must not be disabled. And it is implementation specific, there are multiple ways you can implement the interaction between layers.

And in any case, the RMT maintainer should be the first one to judge this. If the maintainer thinks this enum shouldn't be used this way, let's go for the variable approach. No big deal :), but we should stick to the procedures (Maintainers ack/nack their maintained files unless they can't do it for some reason).

edugrasa:
@dstaesse no problem, no offence :) Just that the NACK sounded a bit strange when it didn't came from the maintainer. Constructive comments on potential improvements are always welcome, agree we don't need to wait for the maintainer for those.

edugrasa:
I've tested in on the VWall without flow control. The fix works and avoids the race condition.

miqueltarzan:
ACK

PR #756, opened 2015-10-02
Pristine 1.3 pff default change

edugrasa:
* PFF default policy only matches with destination address if qos_id is 0 (0 is a reserved value that means "any qos-id"). 
* Refactor qos_id_t, aligning it with cep_id_t, port_id_t, etc.

maintainers: @vmaffione @sandervrijders


vmaffione:
Ok for the refactor of qos_id_t. However, I don't think the kernel should be aware of "0" being a special case - I'm pointing out the modification in linux/net/rina/pff-ps-default.c. It's just ok to use "0" as default for the various qos_id fields (e.g. setting to 0 in userspace). In this way you can just let the PFF implementation be unaware of that and it will work the same.

edugrasa:
Hi @vmaffione, no it won't work. For example, in the cherish/urgency multiplexing case we have to consider different qos-ids for scheduling, but they all need to be forwarded the same (until we have a link-state routing algorithm that is qos-aware). So, PDUs will come with different values of qos-id (not 0, needed for scheduling), and the PFF won't find a matching entry for them.

vmaffione:
Ok, you're right on this. However, I think it's better, only for the default policy, just not to check the qos_id of the PDU PCI. Basically, remove that line instead of changing it. In this way you avoid giving 0 a special meaning, and still obtain the "any qos_id" behaviour. What do you think?

edugrasa:
Well, then if you want to have a PFF policy that does check address and qos-id is replicating the full code just for one-line change.. Seems a bit of overkill compared to having a policy that can address both reserving '0' to mean 'any qos-id'. 

It could also be useful if you want to have some PFF entries to have different output ports depending on qos-id for some destination addresses, and have destination addresses through which all qos-id will be forwarded via the same N-1 port. For example:  Let's imagine I have qos-id 3,4,5, output ports 1,2 and destination addresses 24, 35, 58. If we have the '0' meaning any qos-id, you can have a PFF like this <destination address, qos-id, N-1 port>

<24, 0, 1>
<35, 3, 1>
<35, 4, 2>
<35, 5, 2>
<58, 0, 2>

Otherwise you need more entries (since you cannot summarize all qos-ids in a single entry):

<24, 3, 1>
<24, 4, 1>
<24, 5, 1>
<35, 3, 1>
<35, 4, 2>
<35, 5, 2>
<58, 3, 2>
<58, 4, 2>
<58, 5, 2>

vmaffione:
Yes, I've not strong objections to this. It's a matter of opinion, since you can however always write your own policies, if you need something different. By the way, the code replication problem can be circunvented in some way, that's not the real problem.

It's just a matter of opinion of what "PFF default policy" should do. We can do like you propose, so having a more flexible default policy. Or we could just drop the qos-id support in the default policy (what I proposed), so that default PFF policy cannot differentiate between qos-ids.

In any case I'm ok with your approach.

edugrasa:
Agree, I also have no strong opinion, I just like the '0' approach as a default policy a little bit more because it would cover more use cases. 

@sandervrijders could you ACK this PR? (or provide feedback in case there is something you disagree with :))

sandervrijders:
ACK

PR #760, opened 2015-10-07
Pristine 1.3 rbfc

kewinrausch:
Here the rate-based flow control mechanism to be merged with Pristine-1.3.

Changes to:
* DTP and DTCP to include rate based flow control logic.
* Common DTP and DTCP policies to adjust the behavior of default policies.
* Logs, to include the name of the logging procedure(you can remove this if you want, but was useful).
* PCI to include rate based fields in the pdu header.
* PDU to evaluate the size of the PDU.
* SERDES to evaluate the PCI fields sizes. !!! This has hardcoded values now and has to be fixed.

Calling for ACKs:
* @lbergesio for the DTP and DTCP parts.
* @vmaffione for the log part.
* @vmaffione for PCI and PDU part.
* @sandervrijders for the serdes crap part (sorry Sander).


vmaffione:
@kewinrausch  This branch has conflicts, you must solve them before issuingthe pull request

kewinrausch:
Aaargh.... so much frequent commits. 
I guessed it was aligned.

Do I have to do a new PR?

vmaffione:
@kewinrausch, The log change (introducing function name in the printed string) has to go in a separate pull request, and requires other changes also. Please drop that.

kewinrausch:
@vmaffione Well, in the end was only 1 simple field more in the log.h macro. I did no other change anything else nor I saw any additional adjustment for this change.

Changes introduced:
* ~~Logs, to include the name of the logging procedure(you can remove this if you want, but was useful).~~
Now logs are reverted to match the original implementation.

kewinrausch:
:exclamation: After some discussion on the mailing list and the Skype chat I realized that the introduced hook in the DTP ps makes no sense, so i'm going to remove it soon.

kewinrausch:
I aligned (again) my PR to the last version, and tested that the rate-based mechanism work (on VM for the moment, on real nodes soon). I don't believe that there will be problems on such mechanism compared to the last version (before kernel 4.1.10) because there was not any conflict on that side.

Calling again for ACK people:
    @lbergesio for the DTP and DTCP parts.
    @vmaffione for the log part.
    @vmaffione for PCI and PDU part.
    @sandervrijders for the serdes crap part (sorry Sander).

Can we fix (eventually) and merge this please?

linux/net/rina/dtcp-conf-utils.c
```diff
@@ -1217,3 +1217,4 @@ struct policy * dtcp_rtt_estimator(struct dtcp_config * cfg)
         return cfg->rtt_estimator;
 }
 EXPORT_SYMBOL(dtcp_rtt_estimator);
+
```
lbergesio: Sorry, I know this is annoying from my part, but could you remove this blank line?

linux/net/rina/dtp-ps-default.c
```diff
@@ -84,6 +84,7 @@ int
 default_closed_window(struct dtp_ps * ps, struct pdu * pdu)
```
lbergesio: Could you clean unneeded comments in this function?
The if clause for rate_based is really needed?:
```
+	if(dtcp_rate_based_fctrl(dtcp_config_get(dtcp))) {
+		LOG_DBG("rbfc CWQ: %zd, lim: %u", cwq_size(cwq), max_len);
+	}
```
The changes I think are valid here are the check on ```max_len``` and the ASSERT on the overrun policy.

linux/net/rina/dtcp.c
```diff
@@ -747,9 +1125,32 @@ static int rcv_flow_ctl(struct dtcp * dtcp,
...
         ASSERT(pci);
 
-        snd_rt_wind_edge_set(dtcp, pci_control_new_rt_wind_edge(pci));
+        rt = pci_control_sndr_rate(pci);
+        tf = pci_control_time_frame(pci);
+
+        // Is this a window based control flow?
+        if(dtcp_window_based_fctrl(dtcp_config_get(dtcp))) {
+        	snd_rt_wind_edge_set(dtcp, pci_control_new_rt_wind_edge(pci));
+        	dt_sv_window_closed_set(dtcp->parent, false);
```
lbergesio: Why do we need to set this here? (window open)

linux/net/rina/dtp-ps-default.c
```diff
@@ -312,11 +337,14 @@ dtp_ps_default_create(struct rina_component * component)
         ps->priv            = NULL;
 
         ps->transmission_control        = default_transmission_control;
+        ps->rx_control			= NULL;
         ps->closed_window               = default_closed_window;
-        ps->flow_control_overrun        = default_flow_control_overrun;
+        ps->snd_flow_control_overrun    = default_snd_flow_control_overrun;
+        ps->rcv_flow_control_overrun    = NULL;
```
lbergesio: Is this policy really needed?

linux/net/rina/pci.c
```diff
@@ -83,6 +86,8 @@ ssize_t pci_length(const struct pci * pci)
 { return pci ? sizeof(*pci) : -1; }
 EXPORT_SYMBOL(pci_length);
 
+
+
```
lbergesio: These lines should be deleted

linux/net/rina/pci.c
```diff
@@ -490,3 +495,65 @@ seq_num_t pci_control_last_seq_num_rcvd(const struct pci * pci)
...
+		return -1;
+	}
+
+	pci->control.time_frame = frame;
+
+	return 0;
+}
+EXPORT_SYMBOL(pci_control_time_frame_set);
+
```
lbergesio: This line to be deleted

linux/net/rina/dtp-ps-default.c
kewinrausch: These policies were introduced by Miquel if I'm not wrong. 
We can discuss with him about this. The idea behind, if i'm not wrong, was to not block receiving data if the send rate have been fulfilled (and so maintain the mechanism separated).

linux/net/rina/dtcp-conf-utils.c
kewinrausch: Sure, committing right now.

linux/net/rina/pdu.c
```diff
@@ -287,3 +288,9 @@ int pdu_destroy(struct pdu * p)
         return 0;
 }
 EXPORT_SYMBOL(pdu_destroy);
+
+uint_t pdu_length(struct pdu * pdu) {
+	return buffer_length(pdu->buffer) +
+		serdes_pci_size(pci_type(pdu->pci));
+}
+EXPORT_SYMBOL(pdu_length);
```
lbergesio: isnt't this length actually the serialized length? If so, shouldn't it be in serdes.c or pdu-ser.c?

linux/net/rina/dtp-ps-default.c
kewinrausch: That was a rage comment about decrements an unsigned int without checking for overflows (well, it's a valid condition for every variable at the end) . :scream:

I can remove it now, no prob.

About that additional if condition: it was there only for debugging purposes, so I'm removing it right now.

linux/net/rina/dtcp.c
kewinrausch: We need to distinguish when using window- or rate-based FC, and then react to the receiving of the flow control in a separate way. Makes no sense to update the right window edge when we have rate-based FC(which does not care of windows).

If the flow is not aligned here then we can have, in the same dif, two IPCP with a flow emitting at different speed (depending on which node you are looking at). The flow is unique and the endpoints of such flow are maintained aligned on the same rate.

linux/net/rina/pci.c
kewinrausch: Will do... :)

linux/net/rina/pci.c
kewinrausch: Removing that ending extra line too...

linux/net/rina/pdu.c
kewinrausch: Could be, I had some problem with this is serdes.c... 

I'm waiting for @sandervrijders ACK about this, because now serdes_pci_size is hardcoded with some values which are passed in the DIF configuration(addresses length and so on).

The part which depends on the serializer is in fact handled by that serializer procedure, but this procedure has a pdu context (length of the pdu, not of it's data).

linux/net/rina/dtcp.c
lbergesio: I do not get it. The if clause is ok, so distinguishing between window or rate based fc. My question is, why in the case of window base you set ```dt_sv_window_closed_set(dtcp->parent, false);``` (line 1134)

kewinrausch: That code was inherited from previous versions... and removed from you and Eduard the 25th August. Probably left behind due the long pause during end of August, first of September, when your commits were pushed.

I guess the correct part there should be:
```
snd_rt_wind_edge_set(dtcp, pci_control_new_rt_wind_edge(pci));
push_pdus_rmt(dtcp);
```

right?

linux/net/rina/pdu.c
lbergesio: Yes, I have seen now this (
https://github.com/IRATI/stack/pull/760/files#diff-6b1399bf2f445bfa9cd35be0a403e2bfR1126).
I do not think this is ACKable because in the current form
```serdes_pci_size```
will not work if you change any configuration variable (seq num length, qos
leng
th and so on).
In the same way you crated ```serdes_pci_size``` in serdes.c, you should
cahnge it  to use the ```struct dt_cons``` for this. ```serdes_pci_size```
would become and static utility function within serdes.c and
```pdu_length``` in pdu.c should become something like ```ser_pdu_length```
in serdes.c

2015-11-10 13:19 GMT+01:00 Kewin Rausch <notifications@github.com>:

> In linux/net/rina/pdu.c
> <https://github.com/IRATI/stack/pull/760#discussion_r44399697>:
>
> > @@ -287,3 +288,9 @@ int pdu_destroy(struct pdu * p)
> >          return 0;
> >  }
> >  EXPORT_SYMBOL(pdu_destroy);
> > +
> > +uint_t pdu_length(struct pdu * pdu) {
> > +	return buffer_length(pdu->buffer) +
> > +		serdes_pci_size(pci_type(pdu->pci));
> > +}
> > +EXPORT_SYMBOL(pdu_length);
>
> Could be, I had some problem with this is serdes.c...
>
> I'm waiting for @sandervrijders <https://github.com/sandervrijders> ACK
> about this, because now serdes_pci_size is hardcoded with some values which
> are passed in the DIF configuration(addresses length and so on).
>
> The part which depends on the serializer is in fact handled by that
> serializer procedure, but this procedure has a pdu context (length of the
> pdu, not of it's data).
>
> —
> <https://github.com/IRATI/stack/pull/760/files#r44399697>.
>



-- 
Leonardo Bergesio
Distributed Applications and Networks Area - DANA
i2CAT Foundation - www.i2cat.net
C/ Gran Capità 2-4, office 109, Nexus I building
08034 Barcelona, Catalonia (Spain)
T: +34 93 567 9926
F: +34 93 553 2520
W: http://dana.i2cat.net

File: linux/net/rina/dtcp.c (hunk 3 above)
Comment (lbergesio) [inline linux/net/rina/dtcp.c]: You have the ```push_pdus_rmt(dtcp);``` down in line 1149 which is done in both cases and I think it is right. The window based clause should be only ```snd_rt_wind_edge_set(dtcp, pci_control_new_rt_wind_edge(pci));```

**Comment by sandervrijders** [comment]
Don't use C++ style comments.

**Comment by kewinrausch** [comment]
What do you mean with C++ style comments?

File: linux/net/rina/pdu.c (hunk 7 above)
Comment (kewinrausch) [inline linux/net/rina/pdu.c]: Fixed that part after lot of invocations to god (thanks Vincenzo). 

I added a new issue in IRATI stack for the dt constant structure missing fields for rate based pci fields sizes. Now parse_dt_cons in rnl_utils has the following fix, which can be replaced once the support for this field is introduced.

```
/* FIXME: To modify once this field are supported by rnl. */
dt_cons->rate = 4;
dt_cons->frame = 4;
```

**Comment by kewinrausch** [comment]
Main modifications about C++ code comments and pdu_length moved in serdes and depending to dt_cons have be done. Can we move forward with the verification and merge?  :grinning:

**Comment by lbergesio** [comment]
@kewinrausch  What do you mean? I see a change in  rinad/src/ipcm/configuration.cc for the rate but not for the time and RATE_LEN and TIME_LEN are still defined in serdes.c right?

**Comment by kewinrausch** [comment]
@lbergesio Modification of rinad/src/ipcm/configuration.cc are just to handle:
 ```
"rateBasedConfig" : 
{
        "sendingRate" : XXX,
        "timePeriod" : YYY
}, 
```
and so brings DIF rate and time configuration of the desired QoS cube. The name is ```rate_based_config__``` but it stores both the values (by the way, i did not choose that name).

What is missing (see the issue https://github.com/IRATI/stack/issues/790) is the length of the fields in the pci for rate based case, as we have address Length, cep Length, etc... 

In serdes they were defined as:
```
/* FIXME: To be also defined in dt_cons when rate based fc is added */
#define RATE_LEN      4
#define TIME_LEN      4
```
and so 32 bits fields in the pdu header. Now I fixed serdes to get pdu and pci sizes depending on the given ```dt_cons``` structure (and no more by pre-preocessor defines), but the structure itself does not have the rate-based FC values correctly passed by user space.

So rnl_utils.c ```parse_dt_cons()``` actually has been fixed as:
```
/* FIXME: To modify once this field are supported by rnl. */
dt_cons->rate = 4;
dt_cons->frame = 4;
```

lbergesio:
@kewinrausch please merge and test my PR: https://github.com/kewinrausch/stack/pull/1

kewinrausch:
Done. I'm testing it too on my VMs.

kewinrausch:
@lbergesio I performed some tests with both Window- and Rate-based FC and everything seems good.

linux/net/rina/serdes.c
```diff
@@ -1065,3 +1101,82 @@ struct pdu * pdu_deserialize_ni(const struct serdes * instance,
                                 struct crypto_blkcipher * blkcipher)
 { return pdu_deserialize_gfp(GFP_ATOMIC, instance, pdu, dup_conf, blkcipher); }
 EXPORT_SYMBOL(pdu_deserialize_ni);
+
+int serdes_pci_size(pdu_type_t type, struct dt_cons * c) {
```
lbergesio: this function should be static, remove the EXPORT_SYMBOL and the prototype on serdes.h

linux/net/rina/pdu.c
```diff
@@ -29,6 +29,7 @@
 #include "debug.h"
 #include "pdu.h"
 #include "pci.h"
+#include "serdes.h"
```
lbergesio: this include is not needed, right?

edugrasa:
BTW @kewinrausch for next commits please use the commit message format that we agreed during the Oslo meeting ;)

linux/net/rina/serdes.c
kewinrausch: Do you really want to remove the possibility to know just how much a pci weight?

linux/net/rina/pdu.c
kewinrausch: No, it's no more necessary. I just forgot it. :(
Going to remove this.

kewinrausch:
@edugrasa: Guess I forgot about this! :scream:
Do we have a link on the wiki about the commit format, so I can read it again?

linux/net/rina/serdes.c
lbergesio: My point is that in the code as it is now, it is not needed, it is only used by ```serdes_pdu_size```. If we eventually need to access it from somewhere else, it can then be exported.

edugrasa:
http://chris.beams.io/posts/git-commit/

The subject must have this format

librina: component: <subjetct>     or
rinad: component: <subject>  or 
kernel: component: <subject>

e.g.

kernel: rmt: improved send pdu workflow

linux/net/rina/serdes.c
kewinrausch: Going to fix it then :)

kewinrausch:
I'm testing the last version of pristine-1.3-rbfc on vWall too, and seems to behave correctly. I guess that Leonardo modifications also solve the issue https://github.com/IRATI/stack/issues/790.

lbergesio:
@kewinrausch , yes they do.

I have reviewed the code and do some minor testing but did not test it in more challenging scenarios (at least normal over normal over ethernet, both Win and rate base fc with and without rtx). It would be nice if someone that has the scenario in the wall or some other environment can try it out. After that, I would be ok with the PR, maybe cleaning up the commits... but I will leave that to @vmaffione to valorate.

kewinrausch:
I logged an interesting problem on the vWall, or so it seems. This can also be the source of the problems that I bound to ethtool, and that probably was not really coming from ethtool.

I setup the experiment (always same topology) to go at 80Mb/s, and run the client-server in order to register the effective performances between the nodes.

On my client I got:
```
15153(1447679364)#traffic-generator (INFO): sent statistics:   1204545 SDUs,    602272500 bytes in  59999566 us, 80.3036 Mb/s
```

while the server computes:
```
15037(1447679365)#traffic-generator (INFO): Port    3:     13261 SDUs,      6630500 bytes in     60152 ms, 0.8818 Mb/s
```

I spent all the morning looking for the problem (I saw that no-flow control communication was working less or more good, with 400Mb/s performances), and I was thinking about a problem in the rate based implementation too.

But after seeing that the client respect the rate, and the intermediate node does not report any problem, I looked at more basic problems. So after a logging of the networking interfaces I got:

Before experiment (note, the numbers are the packets visible by **ifconfig** statistics):
```
alpha: 0 <- - - - - - - - - -> 0 charlie 0 <- - - - - - - - - -> 0 delta
```

After the experiment:
```
alpha: 1204387 <- - - - - - - - - -> 1204385 charlie 1204384 <- - - - - - - - - -> 13896 delta
```

This is a huuuuuge difference (and respect the proportion of the performances I got), but the intermediate node and the client are sending the right amount of traffic (80Mb/s) while the server is only receiving a very little part of this. 

The problem really **seems to be located in vWall** networks, or so the test say this.

This could also generate the problem of ethtool bad response; it was not ethtool the problem, but the network itself.

edugrasa:
ACK, window-based flow control continues working and rate-based seems to work in basic tests (have not the time to replicate Kewin's advanced tests, but he will find bugs if there still are in rate-based flow control ;)). @vmaffione if you ACK this PR we can proceed to the merge

vmaffione:
ACK

PR #763, opened 2015-10-14
rebase&clean: header cleanage

bernatgaston:
debug.cc contained config.h which was unneeded.
header inclusion in common was doing and unneeded loop (going down
in the folder structure to go up later)

Ack need from @edugrasa


edugrasa:
ACK!

PR #768, opened 2015-10-20
Default policies for kernelspace plugins

vmaffione:
With this patch, plugins writer can specify NULL as kernelspace policy-set methods, to indicate the default policy is required. This is particularly useful with DTCP and DTP.

More checks have been added to the Component Delivery Workflow.
The kernelspace default policy code has been reorganized.

Kernelspace plugins in the root directory have been conveniently updated.


vmaffione:
Once #769 gets acked, this will be also merged.

PR #770, opened 2015-10-22
Add template for external build of userspace plugin

vmaffione:
This patch:
- extends the rinad build system to install the headers required by userspace plugins.
- provides an example userspace plugin in the plugins/ directory

In this first version, the user is required to set the INSTALLDIR environment variable in order to tell the build system where the IRATI installation is.


edugrasa:
ACK, compilation of rinad and the plugin works, merging!

Issue #683, opened 2015-07-14
pristine-1.3 Java bindings crash on using getTimedWait()


mcrotty:
Java VM crash when using getTimedWait()

Stack dump below:

```
Stack: [0x00007f622bfcb000,0x00007f622c0cc000],  sp=0x00007f622c0ca6f0,  free space=1021k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [librina_java.so.0.0.0+0x343b3]  Java_eu_irati_librina_rinaJNI_IPCEventProducerSingleton_1eventTimedWait+0x33

Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
j  eu.irati.librina.rinaJNI.IPCEventProducerSingleton_eventTimedWait(JLeu/irati/librina/IPCEventProducerSingleton;II)Leu/irati/librina/IPCEvent;+0
j  eu.irati.librina.IPCEventProducerSingleton.eventTimedWait(II)Leu/irati/librina/IPCEvent;+7
j  rina.apps.manager.server.BasicServer.execute()V+31
j  rina.apps.manager.Main.main([Ljava/lang/String;)V+282
v  ~StubRoutines::call_stub
```

edugrasa:
Fixed by #775

PR #775, opened 2015-10-23
Fix for #683, and related Java VM crashes

mcrotty:
1. Fixes a NPE, passes events to the Java application
2. Fixes wrapping code generating calls to non-existent Java exceptions. (segfaults) 

@bernatgaston

### DIFF
--- a/librina/wrap/librina.i
+++ b/librina/wrap/librina.i
@@ -165,6 +165,7 @@ SWIG_JAVABODY_TYPEWRAPPER(public, public, public, SWIGTYPE)
     jenv->ThrowNew(excep, $1.what());
   return $null;
 }
+/*
 %typemap(throws, throws="eu.irati.librina.AllocateFlowException") rina::AllocateFlowException {
   jclass excep = jenv->FindClass("eu/irati/librina/AllocateFlowException");
   if (excep)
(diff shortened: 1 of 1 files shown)


librina/wrap/librina.i
```diff
@@ -402,21 +434,22 @@ SWIG_JAVABODY_TYPEWRAPPER(public, public, public, SWIGTYPE)
 DOWNCAST_IPC_EVENT_CONSUMER(eventWait);
 DOWNCAST_IPC_EVENT_CONSUMER(eventPoll);
 DOWNCAST_IPC_EVENT_CONSUMER(eventTimedWait);
-
-%module(directors="1") cdapcallbackjava
+
+%module(directors="1") cdapcallbackjava
 
 %{
+#include <iostream>
```
bernatgaston: Why? Can't be importedwith the import directive?

bernatgaston:
It's an ack from my side but waiting for the explanation on the comment in line 441

librina/wrap/librina.i
mcrotty: We dont want swig to wrap the data-types in iostream, just use the cerr object in the generated native code, to print a warning, if a AP bound IPCEvent has no Java equivalent.

vmaffione:
@bernatgaston , can we merge this or you need more time?

bernatgaston:
Yes please merge, as I said is an ack from my side

PR #780, opened 2015-10-28
Port-id based selection of DTP/DTCP policy sets

vmaffione:
- Move DTP/DTCP policy-set selection and set-policy-set-parameter machinery from efcp.c to normal-ipcp.c implementation. This was necessary because only the normal IPCP knows about the mapping between cep-ids and port-ids.
- Allow DTP/DTCP policy-sets (and policy-set-specific parameters) to be addressed using the port-id rather than the cep-id. This is convenient because the cep-ids can change during the life of a flow, while port-ids don't.
- Fix the policy-set tracking system so that DTP/DTCP policy-sets are associated to flows instead fo being associated to IPCPs (which does not make sense)


lbergesio:
ACK
PSs for DTP and DTCP changed using the IPCM console with a path like: ```efcp.<port-id>.<dtp/dtcp>``` and the configuration file.

BUT one remark regarding point 2: Although we have not come up with a use case for this yet, conceptually the connection and therefore the efcp, dtp and dtcp are identified by the cep-id; and also RINA allows to have several connections/efcp instances (different cep-ids) supporting the same flow (port-id). So, from an implementation point of view this is ok since we only support 1 efcp per flow (flow->active), but conceptually it is not. I think that point 1 is a consequence of 2, so this comment also affects to it.

edugrasa:
Agree with the remark of @lbergesio, but I think the PR is worth merging anyway (as Leo says it is more of a conceptual discussion and we won't have problems in the implementation by associating the DTP/DTCP policies to a port-id instead of a cep-id) and discuss it in the future. So, merging!

vmaffione:
@lbergesio @edugrasa, I know, connections are identified by cep-ids, and one flow is made of multiple connections (one at a time).
Point here was to make DTP/DTCP policy-set selection more practical: port-id as less "ephemeral" than cep-ids - even if in our implementation it's the same, since there is always one-to-one mapping.

So I think it's easier for the administrator/management system to change DTP/DTCP policy sets by specifying the port-ids in place of the cep-ids. This change is not intended to be "conceptual" (e.g. change the model), just practical from the policy-set selection point of view.

lbergesio:
As you said a flow can be supported by multiple connections. But nothing says it has to be just one at a time, that is my point  (we did not find a use case for that though). From our implementation standpoint it is ok because we do just support 1 connection per flow.

Issue #749, opened 2015-09-21
Upgrade the IPCP daemon to cdapv2 and ribv2


edugrasa:
Upgrade the IPCP Daemon so that it uses the CDAPv2 and the RIBv2 libraries (which are improved versions of the CDAP and RIB libraries the IPCP Daemon currently uses), currently already used by the Management Agent. Once done, deprecate the old CDAP and RIB libraries.

PR #787, opened 2015-11-04
Pristine 1.4 749: ported IPCP to CDAPv2 and RIBv2

edugrasa:
Ported IPCP Daemon to CDAPv2 and RIBv2, removing the old CDAP and RIB implementation files (#749). In doing so the design and implementation of CDAPv2 and RIBv2 has also been improved, reducing the use of dynamic memory allocation and supporting use cases that had not been tried with the MA (authentication messages, A-data messages, CDAP messages forwarded by the IPCM).

All the libraries and modules that were relying on the old CDAP and RIB implementations have been updated. I've tested with 3 IPCPs, normal over normal, using all the authentication policies and encryption.

Maintainers:
@bernatgaston 
@vmaffione 
@msune


bernatgaston:
ack

vmaffione:
Usual tests are ok

PR #789, opened 2015-11-10
Pristine 1.3 sdu protection

olichtne:
This pull request separates the Proof of Concept implementation of SDU Protection from the RMT and SerDes into it's own modules, following the example of how other IPCP components are implemented.

Tested with 2 and 3 machines, with SDU Protection enabled and disabled.

ACKs required from:
@vmaffione 
@miqueltarzan 
@sandervrijders


sandervrijders:
ACK for serdes. 2 points though:

- We named SDU protection DU protection in the stack originally, since it is on the layer boundary. So you could discuss whether it is protecting the SDU or the PDU. But sdu-protection.[ch] is fine for me too. 

- The "common" policy was dropped by @vmaffione's latest changes. If you set the policy to NULL it will select the default policy for that function.

olichtne:
Yeah, I've named it sdu-protection because it's called like that everywhere else so this is makes it more consistent...

For the second point, I've seen that the default/common distinction was being removed in other modules but didn't get to doint that here yet.

vmaffione:
I would deprecate this PR in favour of #791, which includes #789 and add some missing bits.

olichtne:
Ok, I can close this PR then?

vmaffione:
Yes, closing it.

PR #791, opened 2015-11-14
Integrate SDU protection into SDK

vmaffione:
This builds on #789, and adds some SDK-specific support (component delivery workflow) in addition to some minor fixes.

However, something is still missing: @edugrasa @olichtne , where in the DIF template should the name/version of the SDUP policy set be specified? I see security profiles and various policies, but couldn't find a way to map them to the sdup-ps defined in this policy set. Do you have some hint, suggestion about that? what do you guys think?


edugrasa:
The configuration isalready in place (all the policies, versions and
parameters are specified in the auth sdu protection profiles.) Right now
for each profile you can specify the encryption policy, TTL and CRC
separately. Do you want to set the sdu-ps as a join set of crc, ttl and
encryption policies (so less granularity?) TTL, CRC and encryption seem to
be independent from each other, so why bundling them together? Moreover,
the parameters of the encryption policy have nothing to do with the
parameters of TTL.. mixing them in a single sdu-ps configuration would be
extrange I think...

vmaffione:
No, I don't want to have less granularity, just wanted to understand better how the configuration works, and your explanation is a good start, thanks! Currently, this PR proposes a single policy set for all the tree parts, but from this discussion it seems clear that is convenient to split them. So three policy sets.

olichtne:
This is the relevant part of the DIF template that I used for testing (SDU Protection enabled):
```json
"authSDUProtProfiles" : {
    "default" : {
        "authPolicy" : {
            "name" : "PSOC_authentication-ssh2",
            "version" : "1",
            "parameters" : [ {
                "name" : "keyExchangeAlg",
                "value" : "EDH"
            }, {
                "name" : "keystore",
                "value" : "/home/olichtne/irati/private.pem"
            }, {
                "name" : "keystorePass",
                "value" : "test"
            } ]
        },
        "encryptPolicy" : {
            "name" : "default",
            "version" : "1",
            "parameters" : [ {
                "name" : "encryptAlg",
                "value" : "AES128"
            }, {
                "name" : "macAlg",
                "value" : "SHA1"
            }, {
                "name" : "compressAlg",
                "value" : "default"
            } ]
        },
        "TTLPolicy" : {
            "name" : "default",
            "version" : "1",
            "parameters" : [ {
                "name" : "initialValue",
                "value" : "50"
            } ]
        },
        "ErrorCheckPolicy" : {
            "name" : "CRC32",
            "version" : "1"
        }
    }
}
```

It should be the same as what was used for configuring the PoC implementation. Regarding the kernel implementation of this configuration, I plan on refactoring this, once I modify the userspace to support the remaining policies.

edugrasa:
Hi, I'm checking the code, looks ok except for two things:
* Separating the single sdu_ps into three separate Policy sets
* Removing sdu-ps-common (only default should be there, this was agreed 2-3 weeks ago, so it is normal that you didn't know :))

I'm going to address the second one in a commit (it's just moving code around) and then think a bit more about the first one.

vmaffione:
Agree. I was going to do the second point, but please go on!

edugrasa:
Just made a quick test, doesn't work for the normal over normal over eth. use case.. Debugging

olichtne:
With SDU Protection enabled or disabled?

edugrasa:
I've just issued a PR to Vincenzo with sdu ps-common removed (that code is in ps-default now). After checking the sdup code more in depth, I've seen that the sdup code does some configuration parsing that belongs to the policy sets. I'll fix that also.

But first I think we should remove the sdu-ps and have 3 policy sets instead:
* sdup-ttl-ps, with the following operations (from current sdu-ps):
** sdup_set_lifetime_limit_policy, sdup_get_lifetime_limit_policy, sdup_dec_check_lifetime_limit_policy
* sdup-crc-ps, with the following operations (from current sdu-ps):
** sdup_add_error_check_policy, sdup_check_error_check_policy
* sdup-encrypt-ps, with the following operations (from current sdu-ps):
** sdup_encrypt_policy, sdup_decrypt_policy, sdup_add_hmac, sdup_verify_hmac

As you may notice the new encrypt would perform the current compress, add padding and encrypt operations. This is because I don't think they need to be exposed via the PS interface as different functions because they are part of the same policy set and the "protect" method in the sdup.c calls them in this order always (compress, add padding, encrypt). Same thing for the unprotect side, the new decrypt operation will perform the old decrypt, remove_padding, decompress operations.

@olichtne, @vmaffione  if you agree I'll implement these changes (should just be moving code around also, and changing some glue code).

vmaffione:
@edugrasa , that's ok it's what must be done. I would then model The SDU protection module as having three submodules (for the purpose of the component delivery workflow) +1

edugrasa:
ok, so that means that I also need to create sdup-ttl.h/c, sdup-errorcheck.h/c and sdup-encrypt.h/c ?
(same way the RMT has pff.h/c with its own policy set?)

Maybe we can discuss the details in Skype :)

olichtne:
agreed

edugrasa:
Re-tested everything again, works fine. Merging!

PR #792, opened 2015-11-18
pristine-1.4: personality support removed

lbergesio:
Personalities support has been deprecated. The default personality (only one ever used) has
been destroyed. The core initialization now calls kipcm_init() which is the
refactored kipcm_create. The default PS plugin remains as a separete LKM.

This is also a first step towards exporting the system state info to sysfs. After this PR rina root dir will look llike this:
```
sys/
└── rina/
    ├── ipcp-factories/
    │   ├── normal-ipc
    │   └── shim-eth-vlan
    ├── ipcps/
    └── version
```


edugrasa:
ACK, tested with 3 nodes, normal over normal over eth, all the authentication policies and encryption.

PR #797, opened 2015-11-23
Cherish urgency policy set plugin

miqueltarzan:
kernel: plugins: cherish: Add new policy set for cherish urgency.

Maintainers: @vmaffione


edugrasa:
ACK, tested it and verified it "works". Still need to do more experimentation and improve behaviour, but will be done in pristine-1.4.

PR #800, opened 2015-11-25
Pristine 1.4 Changes in RMT  to solve write problem

lbergesio:
Concurrent writes are not longer allowed by means of a new boolean state variable in the n1_port. This should solve the problem with the pending_sdu being lost.

RMT PS simplified due to changes in RMT mechanism.

Some FIXMEs added, but this PR is blocking.


edugrasa:
ACK, tested it, the behavior of the RMT now is also more stable (in window-based flow control when the credit increases performance increases as well, with 200 credit the goodput was 900 Mbps -back to back- in XEN VMs).

PR #801, opened 2015-11-26
New Flow Allocator Policy: Round Robin

miqueltarzan:
In order to test the Cherish/Urgency scheduling introduced in the RMT, a policy to select different QoS cubes is required. The RoundRobin policy chooses the QoS cube in a Round Robin manner.


edugrasa:
ACK, tested and verified expected behaviour (choose different qoscube for each flow allocation in a RoundRobin fashion).

PR #802, opened 2015-11-26
New echo-time operation mode: flood mode

miqueltarzan:
In order to estimate the RTT, a new mode in echo-time operation has been introduced: flood mode. It consists of sending multiple pings at once, and averaging the RTT for each ping.


edugrasa:
ACK, fixed a couple of problems (termination of the client in case SDUs get lost and getting the rate from the appropriate input parameter), and verified functionality.

PR #806, opened 2015-11-27
Introduced prioritization of layer mgmt. PDUs in RMT default policy

edugrasa:
Now layer mgmgt PDUs are put in a dedicated queue (with no max size) and dequeued first always.
It doesn't completely solve the problem for multiple layers, but we can't do better until we have separate layer management and data transfer flows.

Maintainers: @lbergesio


lbergesio:
ACK

PR #807, opened 2015-12-03
Separated application to DIF mappings from main config file

edugrasa:
Now the mappings are in a separate file (da.map), which can be modified will the IPCM is running so that the application to DIF mappings can be updated during the experiment lifetime.

Maintainers: @vmaffione, @bernatgaston

### DIFF
--- a/rinad/etc/Makefile.am
+++ b/rinad/etc/Makefile.am
@@ -33,7 +33,7 @@ ipcmanager.conf: ipcmanager.conf.in \
 	chmod a-w $@.tmp
 	mv $@.tmp $@
 
-CLEANFILES    += ipcmanager.conf default.dif shim-eth-vlan.dif
-BUILT_SOURCES += ipcmanager.conf default.dif shim-eth-vlan.dif
-EXTRA_DIST    += ipcmanager.conf.in default.dif shim-eth-vlan.dif
-sysconf_DATA  += ipcmanager.conf default.dif shim-eth-vlan.dif
+CLEANFILES    += ipcmanager.conf default.dif shim-eth-vlan.dif da.map
+BUILT_SOURCES += ipcmanager.conf default.dif shim-eth-vlan.dif da.map
+EXTRA_DIST    += ipcmanager.conf.in default.dif shim-eth-vlan.dif da.map
+sysconf_DATA  += ipcmanager.conf default.dif shim-eth-vlan.dif da.map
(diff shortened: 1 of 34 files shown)


vmaffione:
What is the point of this?

edugrasa:
Allow you to change the applicarion to DIF mappings without having to
restart the ipcm, which is quite handy when experimenting and when
debugging the stack

vmaffione:
Yeah, I think you've missed my updated comment, sorry about that :)

2015-12-05 21:31 GMT+01:00 Eduard Grasa <notifications@github.com>:

> Allow you to change the applicarion to DIF mappings without having to
> restart the ipcm, which is quite handy when experimenting and when
> debugging the stack
> El 05/12/2015 19:58, "Vincenzo Maffione" <notifications@github.com>
> escribió:
>
> > What is the point of this?
> >
> > —
> > <https://github.com/IRATI/stack/pull/807#issuecomment-162235055>.
> >
>
> —
> <https://github.com/IRATI/stack/pull/807#issuecomment-162243663>.
>



-- 
Vincenzo Maffione

edugrasa:
So you say each application should be aware of what DIF to use? Applications shouldn't necessarily need to be aware of what DIF to use (they can if they want, but is not a requirement). The function in RINA that does the application to DIF mapping is the DIF Allocator, which is not implemented yet as you know. What we currently have is a kind of "static" DIF allocator which is basically the information contained in the application to DIF mappings. I've just made it more usable by separating it from the configuration file and allow it to be updated without restarting the IPCM.

I wouldn't remove this function (if you don't want it just leave the file empty or don't even provide the da.map file, that's another reason for separating it from the main config file). We will need to implement a proper DIF Allocator soon.

vmaffione:
No, I'm not saying that application must necessarily be aware of that, or chose the DIF. Nor I am against this functionality (although I still have to test it).

I was questioning about the need of having applicationToDIFMappings (it's there from the beginning), that is used when applications don't specify the DIF (when registering or requesting a flow). If that static mapping is meant to  be an approximation of the DIF allocator, then it kind of makes sense - as a temporary solution - I was missing this piece.
Otherwise, I wouldn't see why, when a DIF is not specified, the IPCM could just chose one of the available DIF, without the need of a static mapping.

bernatgaston:
Tests not running.

vmaffione:
Working to me --> ACK

edugrasa:
@bernatgaston what tests are not running?

bernatgaston:
ack

PR #809, opened 2015-12-10
Free users of CDAP Provider from having to generate invoke_ids if they don't want to do so

edugrasa:
maintainers: @bernatgaston


bernatgaston:
ack

PR #810, opened 2015-12-11
Pristine 1.3 ECMP implementation

javiergll:
This multipath implementation includes:
- Kernel ECMP forwarding policy using hash-threshold algorithm (inside plugins folder, plugin name "pff-multipath")
- ECMP routing configuration option in link-state default routing policy (code in rinad/src/ipcp/plugins/default/routing-ps.cc)

The ECMP routing can be selected in the dif config file under routing configuration section using:
`{ 
  "name"  : "routingAlgorithm",
  "value" : "ECMPDijkstra"
}`


edugrasa:
There are some files that must be removed from this pull request: 

* check-for-patch.sh 
* kpatch/ilab.t.vlan.transparency.patch
* plugins/multipath/.gitignore
* prepare-vlan-generic.sh

plugins/multipath/mp-plugin-ps.c
```diff
@@ -0,0 +1,73 @@
...
+#include <linux/export.h>
+#include <linux/module.h>
+#include <linux/string.h>
+
+#define RINA_PREFIX "pff-multipath"
+#define RINA_PFF_MULTIPATH_NAME "multipath"
+
+#include "logs.h"
+#include "rds/rmem.h"
+#include "rmt-ps.h"
```
lbergesio: This include can be removed. It would be probably good to check if there are some others

plugins/multipath/pft-ps-multipath.c
```diff
@@ -0,0 +1,737 @@
+/*
```
lbergesio: It would be nice if this file is renamed as pff-ps-multipath.c

edugrasa:
Why was the mode of this file changed?
rinad/Makefile.am 100644 → 100755

edugrasa:
ACK, tested with usual scenario.

PR #816, opened 2015-12-14
Pristine 1.3 dif configurations

bernatgaston:
MA-Manager included DIF configurations …
The MA and the Manager have been modified to include DIF configurations
in the IPC process creation (assignment phase). Now, the MA does not use
a template with the dif configuration but receives all the configurations
from the manager including the DIF configurations but also the neighbors
to enroll.

To this intent, several files have been modified to achieve:
1. Cretation (or reuse) of the configuration structures
2. Modification of the assign to dif process to allow not only a template
but also a direct configuration call
3. Modification of the MA functionalities in the IPC process creation
4. Creation of the needed encoders.

Need ack from @edugrasa @vmaffione


bernatgaston:
Name of the branch is 1.3 but is actually for 1.4 and sync with 1.4

edugrasa:
ACK, tested with normal over normal over eth, with and without encryption.

vmaffione:
Most of changes for reformatting. Most of the reformatting is also not needed.

Reformatting, if necessary, should go in separate pull request. In this case, however, I guess you didn't want to format. Are you using an editor that automatically formats code?

In this way is not possible for me to see what code really changed. Can you isolate just the changes relative to this PR and remove all the reformatting?

bernatgaston:
I had a problem with a couple of files with the reformatting (application.cc basically) which I can not reset since they are mix with other not formmatting changes. Eduard has helped me to check even the formatted files. The problem in my editor is fixed and it should not repeat again. However, the only file you are maintiaing is the resource-allocator.cc and it has not been reformatted so please if you could take a look there I would appreciate. I need this PR to send it to @mcrotty so he can work adapting the manager to the new system.

vmaffione:
ok!

vmaffione:
Usual tests ok, ACK also for the trivial changes to resource-allocator.cc

PR #822, opened 2015-12-21
plugins: red: check return value of rmt_config_get()

vmaffione:
Also do correct cleanup on error conditions.

The missing check was leading to kernel crashes when policy-set selection was performed dynamically (select-policy-set).

MAINTAINERS:
@lbergesio


lbergesio:
ack

PR #824, opened 2015-12-22
plugins: rmt: manage cases when rmt_config_get() returns NULL

vmaffione:
When using the select-policy-set command from IPCM console, rmt_config_get() returns NULL.
Some plugins were crashing because they didn't check the return value.

MAINTAINERS:
M: Leonardo Bergesio <leonardo.bergesio (at) i2cat.net>
	plugins/red/rmt-ps-red.c
	plugins/red/rmt-ps-red.c
M: Miquel Tarzan <miquel.tarzan (at) i2cat.net>
	linux/net/rina/rmt-ps-default.c
	plugins/cherish-urgency/cherish-urgency.c

@lbergesio @miqueltarzan


lbergesio:
ack

miqueltarzan:
ACK

PR #826, opened 2015-12-22
Gracefully closing of the read mng thread

### LINKED ISSUE #825: Gracefully closing of the IPCM
When a normal IPCP is closed, the thread that is reading the managmenet sdus starts failing. Until the ipcp in the userspace is not completely closed, it keeps throwing read errors making the debug difficult and not gracefully.

bernatgaston:
Fixes #825
Added error codes on the kernel side that are caught in the userspace.
When a critical error occurs on the read of the mng sdu in the kernel
it means that the kernel side of the normal IPCP is being closed.
In this case the userspace stops the thread that is polling
the kernel

ACK @lbergesio @edugrasa

### DIFF
--- a/plugins/rdsr-ps/README
+++ b/plugins/rdsr-ps/README
@@ -1,0 +1,45 @@
+Receiver congestion Detection, Sender Reaction policies.
+Author: Kewin Rausch <kewin.rausch@create-net.org>
+
+********************************************************************************
+RESUME
+********************************************************************************
+This set of policies will provide a congestion control set of policies which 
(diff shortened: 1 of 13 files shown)


lbergesio:
ack

edugrasa:
Almost there, but not completely correct imho. The readManagementSdU function in librina-ipc-process should throw two types of exceptions: one indicating an error of which the IPCP can't recover (if the error code is -ESRCH), and another one indicating an error that doesn't meant the IPCP is destroying. Then the thread reading mgmt SDUs would break or continue depending on the exception.

PR #828, opened 2015-12-22
Pristine 1.4 ipcm:Check for wrong version of the configuration file

lbergesio:
The ipcm configuration file now counts on a new setting: configFileVersion. On the other side a file tracked within the repository called ipcm_config_file_version contains the last configuration file version. This file is modified every time something in the config file is added/changed and copied to installationPath/etc/ during rinad installation. 

Now when the ipcm is launched with a custom and maybe outdated configuration file, the ipcm will warn about this and ask to continue or not. The goal of this is to avoid losing time every time a new required setting is added to the configuration file.

Example execution:
```
root@e2-leo-vm1:/home/irati/stack# cd /usr/local/irati/bin/ && ./ipcm -c ../etc/version.conf -l DEBUG
5209(1450826322)#ipcm (DBG): SIGSEGV handler installed successfully
5209(1450826322)#ipcm (DBG): SIGINT handler installed successfully
5209(1450826322)#ipcm (DBG): SIGQUIT handler installed successfully
5209(1450826322)#ipcm (DBG): SIGTERM handler installed successfully
5209(1450826322)#ipcm (DBG): SIGHUP handler installed successfully
5209(1450826322)#ipcm (DBG): SIGPIPE handler installed successfully
5209(1450826322)#ipcm (DBG): Config file is: ../etc/version.conf

Current configuration file version is 1.1.0 but last version is 1.4.1. Do you wanna continue?
(y/n): 
n
Exiting...
```

Maintainer @msune , I would also ask to check it to @vmaffione, @bernatgaston  and @edugrasa


rinad/src/ipcm/configuration.cc
```diff
@@ -395,6 +398,52 @@ void parse_ipc_to_create(const Json::Value          root,
...
+
+        file.open(ss.str().c_str());
+        if (file.fail()) {
+                LOG_ERR("Failed to open last config version file");
+                return;
+        }
+	std::getline(file, last_version);
+        file.close();
+
+	if (f_version.compare(last_version)) {
```
vmaffione: Here it would be better to have f_version != last_version, it's more readable for a C++ programmer (and in general).

rinad/src/ipcm/configuration.cc
```diff
@@ -35,6 +36,8 @@
 #include <librina/json/json.h>
 #include "ipcm.h"
 
+#define CONF_FILE_LAST_VERSION "ipcm_conf_file_version"
```
vmaffione: Very minor comment: CONF_FILE_CUR_VERSION would be a better name.

vmaffione:
Good idea, but I think it's not acceptable that a daemon (which is non-interactive by definition) asks for interactive input from standard input. We should just print a warning and go on.
We could add another option (e.g. --strict-version-check) that if specified turns mismatch into IPCM failure and exit, while by default we just warn and go on.

lbergesio:
yeah.. I know it is a demon and originally I just wanted to print a very noticeable warning and continue... but we do not use it like that. Normally if we do not pay attention to the log we keep going and realize that there was an error in the conf file 2 hours later. Probably the flag option is the best one, I will work on these comments

vmaffione:
Yeah, but I've always used it as a daemon, because it is a daemon, not a
command line utility.
For instance I usually run N virtual machines, using (non-interactive)
scripts to run everything I need, IPCM included.
But it's not only me, users must be allowed to use a daemon as a daemon.


So we agree on using a command line option to force failure on mismatch,
interactivity is really not acceptable, imho.
It could also be an option to always fail on mismatch (without any command
line), I'm ok with both approaches.

2015-12-23 12:26 GMT+01:00 Leonardo Bergesio <notifications@github.com>:

> yeah.. I know it is a demon and originally I just wanted to print a very
> noticeable warning and continue... but we do not use it like that. Normally
> if we do not pay attention to the log we keep going and realize that there
> was an error in the conf file 2 hours later. Probably the flag option is
> the best one, I will work on these comments
>
> —
> <https://github.com/IRATI/stack/pull/828#issuecomment-166866629>.
>



-- 
Vincenzo Maffione

msune:
Always fail in config version mismatch is the way to go IMHO.

lbergesio:
+1, exit with error when version mismatches then

2015-12-23 19:47 GMT+01:00 Marc Suñé <notifications@github.com>:

> Always fail in config version mismatch is the way to go IMHO.
>
> —
> <https://github.com/IRATI/stack/pull/828#issuecomment-166968751>.
>



-- 
Leonardo Bergesio
Distributed Applications and Networks Area - DANA
i2CAT Foundation - www.i2cat.net
C/ Gran Capità 2-4, office 109, Nexus I building
08034 Barcelona, Catalonia (Spain)
T: +34 93 567 9926
F: +34 93 553 2520
W: http://dana.i2cat.net

rinad/src/ipcm/configuration.cc
```diff
@@ -395,6 +398,43 @@ void parse_ipc_to_create(const Json::Value          root,
         }
 }
 
+void check_conf_file_version(const Json::Value& root, const std::string& f)
+{
+	std::string last_version, f_version;
+	char answer = 'x';
```
vmaffione: remove this

vmaffione:
Thinking a bit more about this, I think we should avoid having a ipcm_config_file_version file, and just hardcode the version string in rinad/src/ipcm/configuration.cc.

In fact, the structure of the IPCM configuration file is hardcoded in the very same source file. Therefore it is not useful to have the version in a separate file, it does not add more flexibility.

If you agree I could do the changes, that will make this patch easier.

rinad/src/ipcm/configuration.cc
```diff
@@ -395,6 +398,43 @@ void parse_ipc_to_create(const Json::Value          root,
...
+		ss << ".";
+	} else {
+		ss << f.substr(0, pos);
+	}
+	ss << "/" << CONF_FILE_CUR_VERSION;
+
+        file.open(ss.str().c_str());
+        if (file.fail()) {
+                LOG_ERR("Failed to open last config version file");
+                return;
```
vmaffione: Here we should also fail.
We get rid of this problem if we hardcode version in this source file.

lbergesio:
I had this change already done but not pushed because I was stll not sure.
I am ok with harcoding the version in configuration.cc. I was looking for a
way to make sure that developers modify the ipcm_conf_file_version when
they modify the configuration file. This is not a solution but makes things
easier since the version will be defined in the same file. Anyway, we still
have the same problem with the template in rinad/etc/ipcmanager.conf.in. So
I guess we will have to live with it. Check the following commit, where I
also added some od your comments (well, most of them were already there due
to the different approach). If you want to make any other
changes/improvements just go ahead. I will probably agree and will not be
able to check this again in a couple of days!

2015-12-24 11:55 GMT+01:00 Vincenzo Maffione <notifications@github.com>:

> Thinking a bit more about this, I think we should avoid having a
> ipcm_config_file_version file, and just hardcode the version string in
> rinad/src/ipcm/configuration.cc.
>
> In fact, the structure of the IPCM configuration file is hardcoded in the
> very same source file. Therefore it is not useful to have the version in a
> separate file, it does not add more flexibility.
>
> If you agree I could do the changes, that will make this patch easier.
>
> —
> <https://github.com/IRATI/stack/pull/828#issuecomment-167092388>.
>



-- 
Leonardo Bergesio
Distributed Applications and Networks Area - DANA
i2CAT Foundation - www.i2cat.net
C/ Gran Capità 2-4, office 109, Nexus I building
08034 Barcelona, Catalonia (Spain)
T: +34 93 567 9926
F: +34 93 553 2520
W: http://dana.i2cat.net

vmaffione:
@lbergesio , @edugrasa  for me this is ok now, if you want give a last try we can merge

bernatgaston:
ack from my side

lbergesio:
ack

2015-12-28 11:14 GMT+01:00 Bernat <notifications@github.com>:

> ack from my side
>
> —
> <https://github.com/IRATI/stack/pull/828#issuecomment-167536609>.
>



-- 
Leonardo Bergesio
Distributed Applications and Networks Area - DANA
i2CAT Foundation - www.i2cat.net
C/ Gran Capità 2-4, office 109, Nexus I building
08034 Barcelona, Catalonia (Spain)
T: +34 93 567 9926
F: +34 93 553 2520
W: http://dana.i2cat.net

edugrasa:
Ack!

PR #830, opened 2015-12-30
Gracefully closing of the read mng thread

### LINKED ISSUE #825: Gracefully closing of the IPCM

bernatgaston:
Fixes #825
Added error codes on the kernel side that are caught in the userspace.
When a critical error occurs on the read of the mng sdu in the kernel
it means that the kernel side of the normal IPCP is being closed.
In this case the userspace stops the thread that is polling
the kernel

Comes from https://github.com/IRATI/stack/pull/826 which closed due to an error when uploading the changes. Only need the ack from @edugrasa

### DIFF
--- a/librina/src/ipc-process.cc
+++ b/librina/src/ipc-process.cc
@@ -23,6 +23,7 @@
 
 #include <ostream>
 #include <sstream>
+#include <errno.h>
 
 #define RINA_PREFIX "librina.ipc-process"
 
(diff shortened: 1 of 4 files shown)


edugrasa:
ACK, tested and works

PR #839, opened 2016-01-25
Rib destruction

vicalro:
This PR changes the way some threads read messages so they can be properly joined.
Specifically changes the Console add-on and the Netlink Message Reader.

The method used is to add the sockets to be read inside a select with a timeout. So every time the timeout passes without anything to be read a termination variable can be checked.

The PR also adds the code that deletes all RIB objects to avoid memory leaks.

Please give it a try, and check that you like the current behaviour.

@vmaffione 
@edugrasa 
@bernatgaston 
@lbergesio 
@msune 

Thanks!


edugrasa:
Nice work! Destruction of IPC Processes seems to be working fine, if you can just remove this annoying log ;)

3190(1453809927)#librina.core (DBG): Waiting for message 3190
3185(1453809927)#librina.core (DBG): Waiting for message 1
3201(1453809927)#librina.core (DBG): Waiting for message 3201
3190(1453809929)#librina.core (DBG): Waiting for message 3190
3201(1453809929)#librina.core (DBG): Waiting for message 3201
3185(1453809929)#librina.core (DBG): Waiting for message 1
3190(1453809931)#librina.core (DBG): Waiting for message 3190
3185(1453809931)#librina.core (DBG): Waiting for message 1
3201(1453809931)#librina.core (DBG): Waiting for message 3201
3190(1453809933)#librina.core (DBG): Waiting for message 3190
3201(1453809933)#librina.core (DBG): Waiting for message 3201
3185(1453809933)#librina.core (DBG): Waiting for message 1
3185(1453809935)#librina.core (DBG): Waiting for message 1
3201(1453809935)#librina.core (DBG): Waiting for message 3201

bernatgaston:
Nice work, just one coment, make check fails in librina
-----------------------------------------------
!!!FAILURES!!!
Test Results:
Run:  12   Failures: 1   Errors: 0


1) test: ribBasicOps::testRIBDestruction (F) line: 1276 /home/irati/codi/librina/test/test-rib_v2.cc
assertion failed
- Expression: 0
- Wrong exception thrown during destroy invalid RIB

---------------------------------------------------------------

Could you fix that?

vicalro:
Thanks for checking the PR!

@edugrasa : Annoying log deleted

@bernatgaston : It took me a while(*) but I realized that the error on the test is due to the fact that deleting the RIB does not throw a `eNotImplemented` anymore, because it is actually ... implemented :)

Lots of the tests use the function:
`CPPUNIT_ASSERT_MESSAGE()` 
with a message, and a zero as a second parameter (a.k.a FALSE). This is the assertion that the newly implemented code is hitting. Is that the desired behaviour? Shall we change this second parameter to one every time a feature is implemented?


(*) Even with the flag `--enable-debug` activated, it seems that our `configure.ac` sets both O0 and O2 options so in the end you end up O2 behaviour because is lays after O0.

bernatgaston:
It seems that the changes have to been pushed. Maybe you forgot to push them?

bernatgaston:
ack!

edugrasa:
Ack! merging

PR #850, opened 2016-02-01
Memory leack while unregistering app

bernatgaston:
Valgrind detected a memory leak problem when unregistering app
in rinad. This happened because of exceptions being thrown without
freeing the appropiate memory

ack @edugrasa


edugrasa:
ACK, merging!

PR #853, opened 2016-02-02
MA stuck when closing flows

### LINKED ISSUE #851: MA gets stuck if it has flows allocated
If your enroll a manager to the MA and closes the Manager, the MA can not be closed

bernatgaston:
Fixes #851 . The order of destroying IPCP and addons has been changed.
Now, IPCPs are destroyed before addons to avoid the MA being stuck on
the threads reading these IPCPs. Moreover, the loop of the MA reading
on the flows have been changed to allow exit from this loop.

ack @edugrasa

### DIFF
--- a/rinad/src/ipcm/addons/ma/flowm.cc
+++ b/rinad/src/ipcm/addons/ma/flowm.cc
@@ -515,8 +515,6 @@ FlowManager::~FlowManager()
 	it = workers.begin();
 
 	while (it != workers.end()) {
-		//Stop and join
-		//it->second->stop();
 
 		//Keep next
 		next = it;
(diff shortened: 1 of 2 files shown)


rinad/src/ipcm/addons/ma/flowm.cc
```diff
@@ -357,7 +357,7 @@ void* ActiveWorker::run(void* param)
 			LOG_DBG("Connection stablished between MAD and Manager (port id: %u)", port_id);
 
 			//I/O loop
-			while(true) {
+			while(keep_running) {
```
msune: This sounds strange.

As I think I told you, the agreement with librina's readSDU API was that in case the flow was removed (due to IPCManager destruction => IPCP destruction => Flow destruction), librina should throw an exception.

Is this not happening any more? Is it broken?

bernatgaston: This change was not the cause of the problem but I thought it was useful (the outer loop has this condition) but I didn't remember the exception... you are true. I am repushing

edugrasa:
ACK, merging!

PR #854, opened 2016-02-02
fixes #846

### LINKED ISSUE #846: Pristine-1.4: Multipath PFF policy should handle pointer to 0 case
The case when ```select_entry``` procedure returns a pointer to 0 should be handled in ```mp_next_hop```, otherwise the module will crash trying to read a not valid location (or interrupted if assertion are enabled at compile-time level).

javiergll:
Fixes multipath pff policy null pointer handling

### DIFF
--- a/plugins/multipath/pff-ps-multipath.c
+++ b/plugins/multipath/pff-ps-multipath.c
@@ -525,14 +525,17 @@ static int mp_next_hop(struct pff_ps * ps,
                 return -1;
         }
 
-	list_for_each_entry(pos, &tmp->ports, next) {
-	}
-
         /* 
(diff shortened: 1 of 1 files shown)


edugrasa:
I get this compilation warning

root/openirati/plugins/multipath/pff-ps-multipath.c: In function ‘mp_next_hop’:
/root/openirati/plugins/multipath/pff-ps-multipath.c:489:26: warning: unused variable ‘pos’ [-Wunused-variable]
  struct pft_port_entry * pos;
                          ^
Could you remove that unused variable? Thanks!

edugrasa:
@javiergll Could you remove that line so that we can merge your code? Thanks :)

PR #861, opened 2016-02-06
plugins: add depmod -a instruction in install targets

### LINKED ISSUE #860: Issue depmod command when installing custom plugins
The makefiles in plugins directory don't issue depmod command during installation. This is necessary for the modprobe command (used by the ipcm) to work correctly.

vmaffione:
This makes sure modprobe works as expected after make install.

This fixes #860 

ACK:
@kewinrausch

### DIFF
--- a/plugins/cherish-urgency/Makefile
+++ b/plugins/cherish-urgency/Makefile
@@ -30,6 +30,7 @@ clean:
 install:
 	$(MAKE) -C $(KDIR) M=$$PWD modules_install
 	cp cherish-urgency-plugin.manifest /lib/modules/$(KREL)/extra/
+	depmod -a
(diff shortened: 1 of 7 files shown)


kewinrausch:
I experimented this after having the IRATI kernel installed anew in a vWall real machine (so after, sigh, kernel and user space compilation). Now let the ipcm loading the policies after their compilation and installation works without any intermediate step (which was necessary with older make files). 

**ACKED.**

**Note:** remember to update the dummy policies makefiles too, if someone has to start from that skeleton.

vmaffione:
Thanks! I've not forgot the "dummy", we simply don't have make install there. I will fix this separately.

PR #863, opened 2016-02-09
rinad: ipcm: fixed bug when unregistering apps

edugrasa:
Wrong handling of transactions code in IPCM caused it to terminate when an application was unregistered.

@vmaffione could you test it? It's quick, just register and unregister an application several times and make sure IPCM is still alive.


edugrasa:
Merging since it is a small fix that solves a problem which makes testing very inconvenient (every time an app is unregistered the IPCM dies, requiring a reboot to bring the system back in sync).

vmaffione:
ACK

PR #876, opened 2016-02-29
neighbor lost bug fix

### LINKED ISSUE #864: Pristine-1.4:  Rerouting PDUs of a new/ongoing flow through different link fails
**What is my Issue**
If I disconnect a redundant link between two nodes (by blocking the VLAN) New flow Allocation requests and ongoing flows do not reroute through other links other than the disconnected link. The fact that ongoing flows do not reroute is apparently a known issue but I could not find the github issue, I was told that new flows would reroute but this is not happening for me. When the disconnected link is brought back up before the neighbour_declared_dead event occurs, the network does restore to a state where new flows can be transfered.

**When does it occur**
I have set up a network as [Physical.pdf](https://github.com/IRATI/stack/files/124619/Physical.pdf) and [RINA.pdf](https://github.com/IRATI/stack/files/124620/RINA.pdf), note that these are 4 virtual nodes on one desktop. The configuration files of every virtual node are [here](https://github.com/IRATI/stack/files/124617/Config.zip).

For my issue to occur I disconnect the link between System 1 and System 2. After this I wait a while for the RIBs to update with new hops and start new rina-echo-time instances. System 1 is only possible to communicate with directly connected System 3. The RIB of System 1 shows the disappearance of System 2 (Address 17) but System 4 (Address 19) still remains. System 1 can't communicate with either System 4 or System 3.

**More details**
The address of the ipcp of System 4 also disappears from the RIB of System 1. When trying to send PDUs from System 3 rina-echo-time results in an error and the log of System also indicates some errors.

errors

````
root@rina-echo-time -c 10
24620(1453301930)#librina.logs (DBG): New log level: INFO
24620(1453301930)#librina.nl-manager (INFO): Netlink socket connected to local port 24620 
24620(1453301930)#rina-echo-time (ERR): Failed to allocate a flow
24620(1453301930)#rina-echo-time (ERR): Problems creating flow, exiting
SDUs sent: 0; SDUs received: 0; 0% SDU loss
Minimum RTT: 9.2234e+18 ms; Maximum RTT: 0 ms; Average RTT:0 ms; Standard deviation: -0 ms
````
Next hops Before failure
````
Name: /resalloc/nhopt/key=17-0; Class: NextHopTableEntry; Instance: 100
Value: Destination address: 17; QoS-id: 0; Cost: 1; Next hop addresses: 17/ 

Name: /resalloc/nhopt/key=18-0; Class: NextHopTableEntry; Instance: 101
Value: Destination address: 18; QoS-id: 0; Cost: 1; Next hop addresses: 18/ 

Name: /resalloc/nhopt/key=19-0; Class: NextHopTableEntry; Instance: 102
Value: Destination address: 19; QoS-id: 0; Cost: 1; Next hop addresses: 17/ 
````

Next hops After failure
````
Name: /resalloc/nhopt/key=18-0; Class: NextHopTableEntry; Instance: 155
Value: Destination address: 18; QoS-id: 0; Cost: 1; Next hop addresses: 18/ 

Name: /resalloc/nhopt/key=19-0; Class: NextHopTableEntry; Instance: 156
Value: Destination address: 19; QoS-id: 0; Cost: 1; Next hop addresses: 18
```

**Extra information**

- Debug options are disabled
- The latest commit of the used Pristine branch have the ids 9ce9644 and 1b4927e.
- I have also tried out the scenario that was executed in this [document](http://ict-pristine.eu/wp-content/uploads/2013/12/pristine_d62-use-case-trials-and-business-impact_draft.pdf) with the same configurations. The traffic did not reroute in my experiments and modifying the timers (e.g DeclaredDeadIntervalinMS) did not have an impact on the link failure resiliency.
- Afterwards I have used the multipath plugin developed by Javier Garcia which sent the PDUs through multiple paths. However, when I disconnected the same link as before (System 1 - System 2) the very same issue occurred

bernatgaston:
When a neighbor is lost, the routing was erasing all the FSOs
that had a realtion which such neighbor. It has only to erase
the direct link between curren IPCP and the neighbor and not
any remote link with the address of the neighbor.
Fixes #864

### DIFF
--- a/rinad/src/ipcp/plugins/default/routing-ps.cc
+++ b/rinad/src/ipcp/plugins/default/routing-ps.cc
@@ -1105,7 +1105,8 @@ void FlowStateObjects::deprecateObject(const std::string& fqn,
 	}
 }
 
-void FlowStateObjects::deprecateObjects(unsigned int address,
+void FlowStateObjects::deprecateObjects(unsigned int neigh_address,
+                                        unsigned int address,
 		     	     	        unsigned int max_age)
 {
 	rina::ScopedLock g(lock);
@@ -1113,7 +1114,7 @@ void FlowStateObjects::deprecateObjects(unsigned int address,
 	std::map<std::string, FlowStateObject *>::iterator it;
 	for (it = objects.begin(); it != objects.end();
 			++it) {
-		if (it->second->get_neighboraddress() == address ||
+		if (it->second->get_neighboraddress() == neigh_address &&
 				it->second->get_address() == address) {
 			it->second->deprecateObject(max_age);
 			modified_ = true;
@@ -1428,9 +1429,10 @@ void FlowStateManager::getAllFSOs(std::list<FlowStateObject>& list) const
 	fsos->getAllFSOs(list);
 }
 
-void FlowStateManager::deprecateObjectsNeighbor(unsigned int address)
+void FlowStateManager::deprecateObjectsNeighbor(unsigned int neigh_address,
+                                                unsigned int address)
 {
-	fsos->deprecateObjects(address,
+	fsos->deprecateObjects(neigh_address, address,
 			       maximum_age);
 }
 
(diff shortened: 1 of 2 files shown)


edugrasa:
ACK, tested and works well!

PR #882, opened 2016-03-11
Pristine 1.5 authssh

edugrasa:
Fixed an issue identified by Sarah from Thales: the policy was using a single RSA key pair for both authenticating parties, which doesn't make a lot of sense and is not how SSH operates.  Now is updated so that each IPCP has its own RSA key pair. Therefore, each IPCP needs to know the public keys of the IPCPs it wants to establish an application connection to.

The code is slightly changed, as well as the configuration of this policy. Now the "keystore_path" configuration parameter expects a path to a folder (terminating without "/"). The contents of the folder should be the following files:

* **key** : The RSA private key of the IPCP in PEM format (generated for example with _openssl genrsa -out key 2048_)
* **public_key**: The RSA public key of the IPCP in PEM format (extracted from the private key file, for example with _openssl rsa -in key -pubout > mykey.pub_)
* For each IPCP whose public key is known, a file containing his public key in PEM format. The file *must be named* with the IPC Process application process name. For example, imagine the IPCP has 3 neighbors, then there would be 3 files, each one containing the public key
    * B.IRATI
    * C.IRATI
    * D.IRATI

### DIFF
--- a/librina/include/librina/security-manager.h
+++ b/librina/include/librina/security-manager.h
@@ -62,6 +62,7 @@ public:
 
 	/// get auth_policy
 	virtual cdap_rib::auth_policy_t get_auth_policy(int session_id,
+							const cdap_rib::ep_info_t& peer_ap,
 					   	 	const AuthSDUProtectionProfile& profile) = 0;
 
 	/// initiate the authentication of a remote AE. Any values originated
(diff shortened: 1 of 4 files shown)


edugrasa:
I'm the maintainer, but could @vmaffione take a look if you're not too overloaded? Thanks!

vmaffione:
@edugrasa sure!

vmaffione:
Compilation is OK, regression tests are OK. The patch looks simple and clear.
ACK

PR #885, opened 2016-04-05
Check te pointer to the ipcp before accessing it

### LINKED ISSUE #884: Pristine-1.4 segfault while orinting error in assign_to_dif
While testing the new console I got into a segmentation fault.

These are the 2 commands that I was giving to the console:

`create-ipcp test_console 3 normal-ipc`
`assign-to-dif 3 normal.DIF default.dif`

the error is related to accessing the ipcp pointer when it is still not set
I have already a PR for it

vicalro:
fixes #884

### DIFF
--- a/rinad/src/ipcm/ipcm.cc
+++ b/rinad/src/ipcm/ipcm.cc
@@ -476,9 +476,9 @@ ipcm_res_t IPCManager_::assign_to_dif(
     std::ostringstream ss;
     IPCMIPCProcess* ipcp;
 
-    pre_assign_to_dif(callee, dif_name, ipcp_id, ipcp);
     try
     {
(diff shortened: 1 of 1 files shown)


edugrasa:
I see that the issue is fixed in this function https://github.com/vicalro/irati_stack/blob/d6234bfa5ea6326d334c46ad88bbbb59dd0cef77/rinad/src/ipcm/ipcm.cc#L514, but should also be fixed in this one, right? https://github.com/vicalro/irati_stack/blob/d6234bfa5ea6326d334c46ad88bbbb59dd0cef77/rinad/src/ipcm/ipcm.cc#L471

vicalro:
Yes, indeed.

Another commit already added

edugrasa:
ACK! Merging

Issue #886, opened 2016-04-05
Can't setup multiple shim-tcp-udp IPCP on the same node


jmuchemb:
I try to reproduce [Tutorial 2: DIF over two VLANS](https://github.com/IRATI/stack/wiki/Tutorial-2:-DIF-over-two-VLANS) with shim-tcp-udp, as follows:

* 1st machine (VM0)

  id  | name | type | state | Registered applications | Port-ids of flows provided
  --- | ---- | ---- | ----- | ----------------------- | --------------------------
  1 | vm0.IRATI:1:: | normal-ipc | ASSIGNED TO DIF normal.DIF | - | -
  2 | shim-vm1:1:: | shim-tcp-udp | ASSIGNED TO DIF vm0-vm1 | vm0.IRATI-1-- | 1

      ==> /run/ipcm/shim-vm1.dif <==
      {
        "configParameters": {
          "dirEntry": "1:9:vm1.IRATI0:12:192.168.4.114:1011",
          "expReg": "1:9:vm0.IRATI0:4:1011",
          "hostname": "192.168.4.10"
        },
        "difType": "shim-tcp-udp"
      }
      ==> /run/ipcm/shim-vm2.dif <==
      {
        "configParameters": {
          "dirEntry": "1:9:vm2.IRATI0:12:192.168.4.124:1012",
          "expReg": "1:9:vm0.IRATI0:4:1012",
          "hostname": "192.168.4.10"
        },
        "difType": "shim-tcp-udp"
      }

* 2nd machine (VM1)

  id  | name | type | state | Registered applications | Port-ids of flows provided
  --- | ---- | ---- | ----- | ----------------------- | --------------------------
  1 | vm1.IRATI:1:: | normal-ipc | ASSIGNED TO DIF normal.DIF | - | -
  2 | shim-vm0:1:: | shim-tcp-udp | ASSIGNED TO DIF vm0-vm1 | vm1.IRATI-1-- | 1

      ==> /run/ipcm/shim-vm0.dif <==
      {
        "configParameters": {
          "dirEntry": "1:9:vm0.IRATI0:12:192.168.4.104:1011",
          "expReg": "1:9:vm1.IRATI0:4:1011",
          "hostname": "192.168.4.11"
        },
        "difType": "shim-tcp-udp"
      }
      ==> /run/ipcm/shim-vm2.dif <==
      {
        "configParameters": {
          "dirEntry": "1:9:vm2.IRATI0:12:192.168.4.124:1112",
[...]

Issue #440, opened 2015-02-20
Deadlock in shim-tcp-udp: tcp_udp_sdu_write vs tcp_udp_rcv_worker


kewinrausch:
<a href="https://github.com/lbergesio"><img src="https://avatars.githubusercontent.com/u/1797884?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [lbergesio](https://github.com/lbergesio)**
_Friday Jan 23, 2015 at 18:49 GMT_
_Originally opened as https://github.com/dana-i2cat/irati/issues/437_

----

[  385.163904] 
[  385.164011] =================================
[  385.164011] [ INFO: inconsistent lock state ]
[  385.164011] 3.15.0 #498 Not tainted
[  385.164011] ---------------------------------
[  385.164011] inconsistent {SOFTIRQ-ON-W} -> {IN-SOFTIRQ-W} usage.
[  385.164011] kworker/u17:0/1452 [HC0[0]:SC1[1]:HE0:SE0] takes:
[  385.164011]  (snd_wq_lock){+.?...}, at: [<f80c046b>] tcp_udp_sdu_write+0x4b/0xb0 [shim_tcp_udp]
[  385.164011] {SOFTIRQ-ON-W} state was registered at:
[  385.164011]   [<c108efab>] __lock_acquire+0x46b/0x1a20
[  385.164011]   [<c1090aee>] lock_acquire+0x7e/0x100
[  385.164011]   [<c165fe82>] _raw_spin_lock+0x42/0x70
[  385.164011]   [<f80c2f28>] tcp_udp_write_worker+0x18/0x298 [shim_tcp_udp]
[  385.164011]   [<c105ad9c>] process_one_work+0x19c/0x490
[  385.164011]   [<c105b1b9>] worker_thread+0xf9/0x320
[  385.164011]   [<c10617db>] kthread+0x9b/0xb0
[  385.164011]   [<c1668821>] ret_from_kernel_thread+0x21/0x30
[  385.164011] irq event stamp: 2673675
[  385.164011] hardirqs last  enabled at (2673674): [<c1660136>] _raw_write_unlock_irqrestore+0x36/0x60
[  385.164011] hardirqs last disabled at (2673675): [<c165ffde>] _raw_spin_lock_irqsave+0x1e/0x90
[  385.164011] softirqs last  enabled at (2673546): [<c1545a39>] lock_sock_nested+0x49/0xb0
[...]

Issue #436, opened 2015-02-20
Deadlock in shim-tcp-udp (kernel,sockets,tcp related code) (normal over shim-tcp-udp)


kewinrausch:
<a href="https://github.com/lbergesio"><img src="https://avatars.githubusercontent.com/u/1797884?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [lbergesio](https://github.com/lbergesio)**
_Friday Jan 23, 2015 at 10:53 GMT_
_Originally opened as https://github.com/dana-i2cat/irati/issues/433_

----

[ 4953.732019] BUG: soft lockup - CPU#3 stuck for 22s! [kworker/u17:2:3454]
[ 4953.732019] Modules linked in: shim_tcp_udp nfsd auth_rpcgss oid_registry nfs_acl nfs lockd fscache sunrpc normal_ipcp shim_dummy rina_personality_default 8021q garp stp llc psmouse serio_raw pcspkr i2c_piix4 hid_generic usbmouse usbhid hid floppy
[ 4953.732019] irq event stamp: 433822
[ 4953.732019] hardirqs last  enabled at (433821): [<c1661277>] _raw_spin_unlock_irq+0x27/0x40
[ 4953.732019] hardirqs last disabled at (433822): [<c1660915>] _raw_spin_lock_irq+0x15/0x80
[ 4953.732019] softirqs last  enabled at (433742): [<c15472f2>] release_sock+0x162/0x1c0
[ 4953.732019] softirqs last disabled at (433740): [<c15471be>] release_sock+0x2e/0x1c0
[ 4953.732019] CPU: 0 PID: 3454 Comm: kworker/u17:2 Not tainted 3.15.0 #493
[ 4953.732019] Hardware name: Xen HVM domU, BIOS 4.0.1 05/03/2013
[ 4953.732019] Workqueue: shim-tcp-udp-wwq tcp_udp_write_worker [shim_tcp_udp]
[ 4953.732019] task: f4d3c410 ti: ed7ee000 task.ti: ed7ee000
[ 4953.732019] EIP: 0060:[<c131b3d4>] EFLAGS: 00000202 CPU: 0
[ 4953.732019] EIP is at delay_tsc+0x44/0x70
[ 4953.732019] EAX: 2b23cf18 EBX: 00000001 ECX: 00000000 EDX: 00000aec
[ 4953.732019] ESI: 2b23cef4 EDI: 00000000 EBP: ed7efc6c ESP: ed7efc60
[...]

Issue #424, opened 2015-02-20
General protection fault when sending sdu's over shim-tcp with logs disabled


kewinrausch:
<a href="https://github.com/douwedb26"><img src="https://avatars.githubusercontent.com/u/6333903?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [douwedb26](https://github.com/douwedb26)**
_Wednesday Dec 03, 2014 at 15:40 GMT_
_Originally opened as https://github.com/dana-i2cat/irati/issues/421_

----

Dec  3 14:32:06 y kernel: [  525.029574] rina-kipcm: IPC Manager started. It is listening at NL port-id 1
Dec  3 14:33:01 y kernel: [  580.209250] general protection fault: 0000 [#4] SMP 
Dec  3 14:33:01 y kernel: [  580.209369] Modules linked in: shim_tcp_udp rina_personality_default nfsv3 iscsi_tcp libiscsi_tcp libiscsi scsi_transport_iscsi nfsd auth_rpcgss oid_registry nfs_acl nfs lockd fscache sunrpc psmouse i7core_edac ioatdma edac_core iTCO_wdt gpio_ich iTCO_vendor_support pcspkr lpc_ich serio_raw i2c_i801 tpm_tis hid_generic usbkbd usbmouse usbhid hid igb i2c_algo_bit dca ptp pps_core
Dec  3 14:33:01 y kernel: [  580.210709] CPU: 9 PID: 138 Comm: kworker/u52:0 Not tainted 3.15.0 #11
Dec  3 14:33:01 y kernel: [  580.210754] Hardware name: Supermicro X8DTT/X8DTT, BIOS 2.1b       10/28/2011
Dec  3 14:33:01 y kernel: [  580.210803] Workqueue: shim-tcp-udp tcp_udp_rcv_worker [shim_tcp_udp]
Dec  3 14:33:01 y kernel: [  580.210880] task: ffff8800b461b080 ti: ffff8800b4624000 task.ti: ffff8800b4624000
Dec  3 14:33:01 y kernel: [  580.210927] RIP: 0010:[<ffffffff81183e9d>]  [<ffffffff81183e9d>] __kmalloc+0x8d/0x180
Dec  3 14:33:01 y kernel: [  580.211010] RSP: 0018:ffff8800b4627cb8  EFLAGS: 00010286
[...]

PR #887, opened 2016-04-07
Pristine 1.5 shim TCP, UDP fixes

edugrasa:
Several fixes to the shim DIF over TCP and UDP:

* Fix locking to avoid the problems reported by #424, #436, #440 
* Not sure why it was happening, but #886 seems to be fixed also (at least I could not reproduce it)
* Limited the size of the "outgoing SDUs" queue, which was unbounded and could make the system to run out of kernel memory and crash (happened to me several times). Also added vertical flow control to disable/enable the ports when the queue transitions from not full to full and viceversa.

Maintainers: @sandervrijders. @jmuchemb  could you also give it a try to see if #886 is addressed?

### DIFF
--- a/linux/net/rina/ipcps/shim-tcp-udp.c
+++ b/linux/net/rina/ipcps/shim-tcp-udp.c
@@ -50,6 +50,7 @@
 
 #define CUBE_UNRELIABLE 0
 #define CUBE_RELIABLE   1
+#define SEND_WQ_MAX_SIZE 1000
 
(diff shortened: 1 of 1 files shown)


jmuchemb:
No, this does not fix #886

edugrasa:
So you are still getting the same errors? I've not been able to reproduce them then following the instructions in #886.

sandervrijders:
ACK

PR #893, opened 2016-04-13
rinad: create log directory at install

### LINKED ISSUE #586: pristine-1.3: create the directory for logs if the path doesn't exist upon install
When specifying the path for the config files in the ipcm configuration, the IPCM could check if the path exists and create the directories if necessary instead of reverting to stdout output.

jmuchemb:
This fixes #586
@vmaffione


vmaffione:
ACK

PR #894, opened 2016-04-13
Fixes the parsing of debugging flags in CXXFLAGS

### LINKED ISSUE #890: configure script for librina/rinad/rina-tools does not accept custom CFLAGS
This prevents build from stripping symbols and enable -O2 optimizations.

vicalro:
For librina, rinad and rina-tools
fixes #890

Be aware that this change affects the default compilation of the stack
as --enable-debug is activated by default, and from now on this will
imply -O0, so decreases in performance should be expected.

To get the former performance you should disable --enable-debug
in the configuration scripts

### DIFF
--- a/install-user-from-scratch
+++ b/install-user-from-scratch
@@ -10,7 +10,7 @@
 #
 PREFIX=`echo "$1"|sed -e "s,\/$,,"`
 SRCDIR=`pwd`
-FLAGS="--enable-debug"
+#FLAGS="--enable-debug"
 
 #
 # Do not touch under this line
--- a/librina/configure.ac
+++ b/librina/configure.ac
@@ -90,7 +90,7 @@ AS_IF([test "$want_debug" = "yes"],[
     AX_CHECK_COMPILE_FLAG([ -g ], [
         CXXFLAGS_EXTRA="$CXXFLAGS_EXTRA -g"
     ])
-    AS_IF([`echo $CXXFLAGS|$GREP '\\\-O2'`],[
+    AS_IF([echo "$CXXFLAGS" | "$GREP" '\-O2' > /dev/null],[
         AC_MSG_NOTICE([Patching flags])
         CXXFLAGS="`echo $CXXFLAGS | $SED -e 's|-O2||'`"
         AC_MSG_NOTICE([Flags are now '$CXXFLAGS'])
--- a/rina-tools/configure.ac
+++ b/rina-tools/configure.ac
@@ -93,7 +93,7 @@ AS_IF([test "$want_debug" = "yes"],[
     AX_CHECK_COMPILE_FLAG([ -g ], [
         CXXFLAGS_EXTRA="$CXXFLAGS_EXTRA -g"
     ])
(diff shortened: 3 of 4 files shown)


vicalro:
@edugrasa , @vmaffione Can someone take a look?

Thank you!

vmaffione:
Then it may make sense to have --disable-debug by default. What the other
guysthink?

2016-04-13 14:04 GMT+02:00 vicalro <notifications@github.com>:

> For librina, rinad and rina-tools
> fixes #890 <https://github.com/IRATI/stack/issues/890>
>
> Be aware that this change affects the default compilation of the stack
> as --enable-debug is activated by default, and from now on this will
> imply -O0, so decreases in performance should be expected.
>
> To get the former performance you should disable --enable-debug
> in the configuration scripts
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/IRATI/stack/pull/894
> Commit Summary
>
>    - Fixes the parsing of debugging flags in CXXFLAGS
>
> File Changes
>
>    - *M* librina/configure.ac
>    <https://github.com/IRATI/stack/pull/894/files#diff-0> (2)
>    - *M* rina-tools/configure.ac
>    <https://github.com/IRATI/stack/pull/894/files#diff-1> (2)
>    - *M* rinad/configure.ac
>    <https://github.com/IRATI/stack/pull/894/files#diff-2> (2)
>
> Patch Links:
>
>    - https://github.com/IRATI/stack/pull/894.patch
>    - https://github.com/IRATI/stack/pull/894.diff
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/IRATI/stack/pull/894>
>



-- 
Vincenzo Maffione

vicalro:
I would be fine with --disable-debug :)

vmaffione:
What is the expected result if I

CXXFLAGS="-O3" ./install-user-from-scratch 

?

I see in this case -O0 -g is used (since --enable-debug is used) and -O3 is appended afterwards.

msune:
IMHO, default behaviour has to be O2 or even O3 and maybe without debugging symbols (although this doesn't have performance penalties afaik).

Users don't know (and should not know IMO) the existance of these internal knobs and when to enable/disable. Developers who already know what they are doing, can use ```--enable-debug``` to change to -O0 -g.

edugrasa:
@vicalro what is the conclusion on this one? Thanks :)

vicalro:
I commented out the `--enable-debug` optioin in the script `install-user-from-scratch`. As it is right now default options are `-O2 -g`[*]. Developers can uncomment the line:

`#FLAGS="--enable-debug"`

And happily debug!

[*] In the `configure.ac` files there is a macro named `AC_PROG_CXXCPP` that sets the CXXFLAGS to "-O2 -g". If we want to change the default to "-O3" as @msune proposed we need to change it there.

vmaffione:
ACK from me. I would stick to -O2 now and possibly switch to -O3, if convenient, using a separate PR.

PR #895, opened 2016-04-13
kernel: sysfs: fix CLANG warning on sysfs_create_file return value

lbergesio:
The issue was reported to me by @sandervrijders. The patch has been tested and seems to remove the warning he was getting when compiling with clang (gcc was not doing that afaik).


sandervrijders:
ACK

PR #896, opened 2016-04-14
shim-tcp-udp: add support for IPv6

jmuchemb:
Updated parse_* functions and added support for IPv6


edugrasa:
ACK, tested both over IPv4 (to make sure there are not regressions) and IPv6. Scenario was normal over shim-tcp-udp, both IPv4 and IPv6 work fine!

@sandervrijders if you are fine with the code modifications we can merge this one.

sandervrijders:
Looks good

PR #897, opened 2016-04-17
IPCM: Hide modprobe output when failing to load a module

vmaffione:
The output

````
modprobe: FATAL: could not...
````

used to show up in the logs (e.g. when loading the default userspace plugin), but it is not an error. It is just an indication that the plugin that the IPCM wants to load is not a kernel module.

This patch hides the output using stderr/stdout redirection.


edugrasa:
ACK, tested and works!

PR #899, opened 2016-04-19
rinad: configure: process --disable-java-bindings

vmaffione:
Current configure script ignores the --disable-java-bindings
switch.

This patch aims at making it possible to configure and build rinad without any Java installation on the local machine.


kewinrausch:
**ACK** for the building without Java on a vWall node.

vmaffione:
Thanks, can anyone test that Java bindings are still being built if wanted?

vmaffione:
I've tested it, it seems to be working.

PR #903, opened 2016-04-20
librina: Update C wrapper

sandervrijders:
This updates the C wrapper to have a more C-style IPC API.

Maintainers: @dstaesse


dstaesse:
splendid

librina/wrap/c/src/librina-c.cc
```diff
@@ -0,0 +1,468 @@
...
+
+        it = flows.find(fd);
+        if (it == flows.end())
+                return -1;
+
+        mein_flow = it->second;
+
+        try {
+                ret = ipcManager->readSDU(mein_flow->port_id, buf, count);
+        } catch (IPCException e) {
```
msune: This is a bit dangerous. While wrapping C++ code for C it is _very_ advisable to have a catch all clause (```catch(...){}```). An uncatched exception not deriving from ```IPCException``` (e.g. an STL exception), slipping through the library, will directly force the termination of the C program.

It is also preferable to catch the exception by reference ```catch (IPCException& e){}```.

This also applies to the rest of try-catch clauses

PR #905, opened 2016-04-22
kernel:shim-tcp-udp: fix compilation warning

lbergesio:
Just to avoid the compiler warning about a couple of

"warning: control reaches end of non-void function [-Wreturn-type]"

### DIFF
--- a/linux/net/rina/ipcps/shim-tcp-udp.c
+++ b/linux/net/rina/ipcps/shim-tcp-udp.c
@@ -267,6 +267,7 @@ static unsigned sockaddr_init(union address *         sa,
 		return sizeof sa->in6;
 	}
 	ASSERT(0);
+	return 0;
 }
 
 static unsigned sockaddr_copy(union address * src, union address * dst)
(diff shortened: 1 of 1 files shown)


sandervrijders:
Shouldn't they just have ```static size_t``` as return type instead? And a default case has to be added to the switch case

sandervrijders:
Also, please stop driving your Delorean at more than 88 MPH to create commits in the future.

lbergesio:
I guess changing the return type to `size_t` or `ssize_t` should be ok. The default case in this specific case could be ommited imho. But anyway, I would ask @jmuchemb to tackle this since it is his code.

jmuchemb:
By writing `ASSERT(0)`, I expected the compiler to optimize (when using -O2) by only testing 1 value and treat the other case as a default one.
I've just read the definition of ASSERT and now I understand what's wrong.

Linux offers a macro to do what I had in mind. One solution is to replace all `ASSERT(0)` by `
  unreachable()` and I still have preference for that.

Another solution is to replace `ASSERT(0)` by `BUG()`

About the returned type, I would find overkill for some arch to use big integers whereas we return the size of a small struct. If there's a possible warning here, we can either cast to `unsigned` or change the returned type to `size_t`.

lbergesio:
I have attended @jmuchemb comments. unreachable() was introduced instead of ASSERT(). No warnings. About the return type I agree with @jmuchemb, moreover, both are static fuctions so that their scope should be controlled (it is actuallyis ignored in the actual code).

PR #906, opened 2016-04-22
kernel:shim-eth-vlan: fixed callback when sdu_queue is destroyed

lbergesio:
Original code was:
rfifo_destroy(flow->sdu_queue, (void (*)(void *)) pdu_destroy);
when it should be passing sdu_destroy.

It probably never broke because the sdu_queue was already empty since it is
used only during flow allocation.


sandervrijders:
ACK

PR #908, opened 2016-04-25
kernel: shim-tcp-udp improvements

jmuchemb:
Maintainers: @sandervrijders


edugrasa:
ACK, tested and works, doesn't seem to introduce any regression.

edugrasa:
@sandervrijders could you check if the code is ok? Functional tests are already done..

sandervrijders:
ACK

Issue #909, opened 2016-04-27
Can't make an IPCP enroll to several neighbours


jmuchemb:
With a single shim-tcp-udp to link more than 2 machines, there are cases where making each node a neighbour of all other machines (full-mesh) fail.

Given 3 VMs setup as follows: VM0 -> VM1 -> VM2 where "->" is an enrollment
VM0 -> VM2 then fails (after the TCP connection is established successfully), whereas VM2 -> VM0 succeeds.

Here are logs from VM2:

>  # VM1 -> VM2
> 290(1460111733)#ipcm.flow-alloc (INFO)[flow_allocation_requested_remote]: Arrived request for flow allocation between vm2.IRATI:1:: and Unknown app:::
> 290(1460111733)#ipcm.flow-alloc (INFO)[allocate_flow_response_event_handler]: : Informing IPC process shim2:1:: about flow allocation from application vm2.IRATI:1:: to application Unknown app::: in DIF shim::: [success = 1, port-id = 1]
>
>  # VM0 -> VM2
> 290(1460111735)#ipcm.flow-alloc (INFO)[flow_allocation_requested_remote]: Arrived request for flow allocation between vm2.IRATI:1:: and Unknown app:::
> 290(1460111735)#ipcm.flow-alloc (INFO)[allocate_flow_response_event_handler]: : Informing IPC process shim2:1:: about flow allocation from application vm2.IRATI:1:: to application Unknown app::: in DIF shim::: [success = 0, port-id = -1]

edugrasa:
Just realized what is going on.. one IPC Process is trying to accept two flows from an "Unknown app". The first one will work; for the second one the current default policy checks that the IPCP is not already enrolled to an IPCP with the same name.. and since it is already enrolled to "Unknown app" then it fails. Will issue a fix in short..

edugrasa:
Fixed by #914

PR #913, opened 2016-05-02
riand: ipcp: allow IPCPs to accept flows from multiple unknown IPCPs

edugrasa:
* Fix issue #909 @jmuchemb could you test it? Thanks!

### DIFF
--- a/rinad/src/ipcp/resource-allocator.cc
+++ b/rinad/src/ipcp/resource-allocator.cc
@@ -359,6 +359,10 @@ bool IPCPFlowAcceptor::accept_flow(const rina::FlowRequestEvent& event)
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


edugrasa:
Closing since this introduces new bugs. The real fix is that the shim-tcp-udp passes Ia string identifying the source app as good as it can (source IP address + source port for example, it's all we can have at the shim TCP/UDP level ...)

dstaesse:
yeah, we did that once, we just appended the port_id at the destination (since it's system wide unique). Apparently we didn't do a PR...

edugrasa:
no, wasn't merged, would have been handy ;) I'll then use the port-id as the process instance. (So keep "Unknown app" s process name, use port-id as process instance, as a way to identify different instances of "Unknown app")

PR #915, opened 2016-05-02
kernel: shim-tcp-udp: pass port-id as application process instance of…

### LINKED ISSUE #909: Can't make an IPCP enroll to several neighbours

edugrasa:
Differentiate between multiple instance of "Unknown app" in the shim TCP-UDP. Fixes #909 

Maintainer: @sandervrijders 
@jmuchemb could you test it? Thanks!

### DIFF
--- a/linux/net/rina/ipcps/shim-tcp-udp.c
+++ b/linux/net/rina/ipcps/shim-tcp-udp.c
@@ -1045,7 +1045,7 @@ static int udp_process_msg(struct ipcp_instance_data * data,
         int                         size;
         struct ipcp_instance      * ipcp, * user_ipcp;
         unsigned long		    flags;
-        char			    api_string[10];
+        char			    api_string[12];
 
         LOG_HBEAT;
 
@@ -1181,7 +1181,7 @@ static int udp_process_msg(struct ipcp_instance_data * data,
                 }
 
                 /* FIXME: This sets the name to the server? */
-                if (sprintf(&api_string[0], "%d\n", flow->port_id) < 0){
+                if (sprintf(&api_string[0], "%d", flow->port_id) < 0){
                 	kfa_port_id_release(data->kfa, flow->port_id);
                 	unbind_and_destroy_flow(data, flow);
                         return -1;
@@ -1541,7 +1541,7 @@ static int tcp_process(struct ipcp_instance_data * data, struct socket * sock)
         int                        err;
         struct ipcp_instance     * ipcp, * user_ipcp;
         unsigned long		   flags;
-        char	   		   api_string[10];
+        char	   		   api_string[12];
 
         LOG_HBEAT;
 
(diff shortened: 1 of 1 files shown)


jmuchemb:
Don't you have a memory leak with api_string ? Why not using the stack as a buffer ?

edugrasa:
I don't think there is a memory leak, the memory pointed by api_string is released when the "name" structure is destroyed

edugrasa:
@jmuchemb thanks for pointing at this, fix pushed!

jmuchemb:
This fixes #909 for me.

dstaesse:
we added it to the application name with sprintf, not as the application instance, but this will probably work as well.

linux/net/rina/ipcps/shim-tcp-udp.c
```diff
@@ -1032,6 +1032,22 @@ int send_msg(struct socket *      sock,
         return size;
 }
 
+char * tcp_udp_int_to_string(int data)
+{
+	char * result;
+
+        result = rkmalloc(10, GFP_ATOMIC);
```
dstaesse: 10 characters is usually plenty, but an int may need 12 (one for the - sign, 10 for the number and one for the \0)

linux/net/rina/ipcps/shim-tcp-udp.c
```diff
@@ -1032,6 +1032,22 @@ int send_msg(struct socket *      sock,
...
+        if(!result)
+                return NULL;
+
+        if (sprintf(result, "%d\n", data) < 0) {
```
dstaesse: why the \n?

edugrasa:
@dstaesse Thanks for the feedback! Polished a few details as suggested.

sandervrijders:
ACK. Haven't tested though, but if there are no regressions since @jmuchemb tested it, it can be merged.

PR #919, opened 2016-05-03
Added SIGCHLD handler to IPCM

edugrasa:
Added SIGCHLD handler to IPCM to reap child IPC Process Daemons … (avoid zombies)

Maintainers: @vmaffione

### DIFF
--- a/rinad/src/ipcm/main.cc
+++ b/rinad/src/ipcm/main.cc
@@ -178,27 +178,32 @@ int main(int argc, char * argv[])
 
 	//Configure signal  traps
 	if (signal(SIGSEGV, handler) == SIG_ERR) {
-		LOG_WARN("Could not install SIGSEGV handler!");
+		LOG_ERR("Could not install SIGSEGV handler!");
+		return EXIT_FAILURE;
 	}
         LOG_DBG("SIGSEGV handler installed successfully");
 
 	if (signal(SIGINT, handler) == SIG_ERR) {
 		LOG_ERR("Could not install SIGINT handler!");
+		return EXIT_FAILURE;
 	}
         LOG_DBG("SIGINT handler installed successfully");
 
 	if (signal(SIGQUIT, handler) == SIG_ERR) {
 		LOG_ERR("Could not install SIGQUIT handler!");
+		return EXIT_FAILURE;
 	}
         LOG_DBG("SIGQUIT handler installed successfully");
 
 	if (signal(SIGTERM, handler) == SIG_ERR) {
 		LOG_ERR("Could not install SIGTERM handler!");
+		return EXIT_FAILURE;
 	}
         LOG_DBG("SIGTERM handler installed successfully");
 
 	if (signal(SIGHUP, handler) == SIG_ERR) {
 		LOG_ERR("Could not install SIGHUP handler!");
+		return EXIT_FAILURE;
 	}
         LOG_DBG("SIGHUP handler installed successfully");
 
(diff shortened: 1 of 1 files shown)


rinad/src/ipcm/main.cc
```diff
@@ -203,6 +208,11 @@ int main(int argc, char * argv[])
         }
         LOG_DBG("SIGPIPE handler installed successfully");
 
+	if (signal(SIGCHLD, handler) == SIG_ERR) {
+		LOG_ERR("Could not install SIGCHLD handler!");
+	}
```
vmaffione: add else here

vmaffione:
The idea it's ok, although the init process would do the cleaning up anyway.

edugrasa:
Could you elaborate a bit more? Right now when an IPCP is destroyed (e.g. via the console) the IPCP Daemon process appears as <defunct> until the IPCM is killed. What init process are you referring to?

vmaffione:
Forget what I said, it's late in the night :) I was mixing up with orphan processes.
Yes, this patch is definitely needed to avoid memory leaks.

rinad/src/ipcm/main.cc
edugrasa: I followed the same style as the other ifs that were already in this method. Should all of them be updated then?

edugrasa:
Aaaah, ok :) Should I add "else"s in all that function then?

vmaffione:
Sure, for all the failing signal() that do not cause the IPCM to exit.

rinad/src/ipcm/main.cc
vmaffione: that's not style, that's mistaken : D

edugrasa:
@vmaffione, updated the code to exit on failing signal()

rinad/src/ipcm/main.cc
```diff
@@ -63,6 +64,10 @@ void handler(int signum)
 		case SIGHUP:
 			rinad::IPCManager->stop();
 			break;
+		case SIGCHLD:
+			waitpid(WAIT_ANY, NULL, WNOHANG);
+			LOG_DBG("Child IPC Process Daemon died, removed from process table");
```
vmaffione: I would get the result value, to check if it was successful (>0) or if no children was terminated (0), so that we can emit the correct LOG output. Also, I would call waitpid in a loop, to be more general, and exit the loop when waitpid returns 0 (no more children to reap).

edugrasa:
@vmaffione, update issued, let's see if it addresses your comments

rinad/src/ipcm/main.cc
```diff
@@ -63,6 +64,12 @@ void handler(int signum)
 		case SIGHUP:
 			rinad::IPCManager->stop();
 			break;
+		case SIGCHLD:
+			while (waitpid(WAIT_ANY, NULL, WNOHANG) > 0) {
+				LOG_DBG("Child IPC Process Daemon died, removed from process table");
+			}
+
+			break;
```
vmaffione: Almost ok, but we also need to check if the exit code was -1, to print an error log.

edugrasa:
@vmaffion, updated again

rinad/src/ipcm/main.cc
```diff
@@ -65,10 +67,14 @@ void handler(int signum)
 			rinad::IPCManager->stop();
 			break;
 		case SIGCHLD:
-			while (waitpid(WAIT_ANY, NULL, WNOHANG) > 0) {
-				LOG_DBG("Child IPC Process Daemon died, removed from process table");
+			while (pid = waitpid(WAIT_ANY, NULL, WNOHANG), pid > 0) {
+				LOG_DBG("Child IPC Process Daemon %d died, removed from process table",
+					pid);
```
vmaffione: This is correct, but I personally find the comma operator very bug prone. Anyway,  if you have tested it and it worked it is ok to me.

Issue #902, opened 2016-04-20
Pristine-1.4: Flow accumulates messages that are flushed only at the end.


kewinrausch:
> Whaaat?

If your reaction is like this then you are at the same point as me. This "issue" is visible when you split a flow over multiple Equal cost paths instead of preserving it always on the same one until the end. Actually my policies perform a pick of ECMP ports for a destination in a round-robin style, and so every PDU is sent on a different port (assumed more than one port is available to reach that destination).

I'm using rate based flow with custom policies, which works nicely under normal configuration (all the measurements that has been done until now are correct).

The initial test is done with a classic ECMP policy, which gives me...

**Ping test**

```
krausch@s1:/rping$ ./rping s1 1 dc s4 1 0 10
18384(1461144727)#librina.logs (DBG): New log level: ERR
RTT: 626 us
RTT: 657 us
RTT: 648 us
RTT: 730 us
RTT: 732 us
RTT: 607 us
RTT: 656 us
RTT: 656 us
RTT: 611 us
RTT: 605 us
```

**Bw test**

```
krausch@s4:/rping$ ./rping s4 1 dc --listen
23147(1461145664)#librina.logs (DBG): New log level: ERR
Port 3 allocated for s1-1
Sender:s1, port:3, test:1, recv:6953984, total:6953984, seq:0, ooo: 0, bw:45.876848 Mb/s
Sender:s1, port:3, test:1, recv:12255232, total:19209216, seq:0, ooo: 0, bw:90.109929 Mb/s
Sender:s1, port:3, test:1, recv:12976128, total:32185344, seq:0, ooo: 0, bw:89.177954 Mb/s
Sender:s1, port:3, test:1, recv:13066240, total:45251584, seq:0, ooo: 0, bw:74.633682 Mb/s
Sender:s1, port:3, test:1, recv:16850944, total:62102528, seq:0, ooo: 0, bw:104.722708 Mb/s
Sender:s1, port:3, test:1, recv:15048704, total:77151232, seq:0, ooo: 0, bw:92.888719 Mb/s
Sender:s1, port:3, test:1, recv:11354112, total:88505344, seq:0, ooo: 0, bw:90.473805 Mb/s
Sender:s1, port:3, test:1, recv:13787136, total:102292480, seq:0, ooo: 0, bw:110.407385 Mb/s
Deallocating flow 3...
Sender:s1, port:3, test:1, recv:8740864, total:111033344, seq:0, ooo: 0, bw:69.581440 Mb/s
[...]

PR #920, opened 2016-05-03
RMT: N-1 port enabling/disabling bug fixes

edugrasa:
Fixes a bug that prevented the RMT tasklet from re-scheduling itself in a lot of cases, causing issue #902. Also improves port enable/disable control between shim eth qdisc and the RMT, to avoid the N-1 port from staying disabled forever.

Maintainers: @miqueltarzan. @kewinrausch could you also give it a try?


kewinrausch:
Nice catch Eduard... 
For what I can say now, this solve for sure one of the bug I had, and which we examined some weeks ago (RMT processing it's queue slowly). During the last test I did, my problem was a little different, and the accumulation of the messages (with memory consumption and final flush) occurred even if the RMT queues were at 0 (so all the messages correctly processed by the tasklet).

When the VPN of vWall is restored, I'll give it a try for sure. :)

miqueltarzan:
ACK

edugrasa:
Ok @kewinrausch! I'll merge it since the update is quite critical in case people wants to do experiments, but I won't close issue #902 until you have verified it is fixed.

PR #921, opened 2016-05-06
kernel:rmt:fixed wrong return in n1_port_write_sdu

lbergesio:
n1_port_write_sdu only considered a -EAGAIN coming from the sdu_write ops as
only possible error. If -1 was returned by the op, n1_port_write_sdu was
obviating this and returning the bytes of the sdu supposedly written.


miqueltarzan:
ACK: Tested.

PR #924, opened 2016-05-06
Delegation to a shim IPCP not allowed

### LINKED ISSUE #922: Deal with RIB operations in the shim IPC Processes
Either return an error or some other answer, but improve current behaviour (internal error, log but no feedback to caller)

bernatgaston:
Fixes #922 
Acke @edugrasa


edugrasa:
ACK, merging!

PR #935, opened 2016-05-19
Delegation of objects various bugs fixed

### LINKED ISSUE #932: Delegation objects are not being send to the manager
Delegation objects arribe to the MA but are not being send to the manager

### LINKED ISSUE #934: Double Flags = 0 when returning various RIB Objects
When scope is more than 1, and some of the targetted rib objects are delegated, more than one RIB Object has FLags = 0  (meaning end of the sequence of RIB Obejcts) which produces a change on the state of the cdap session and hence, no more rib obejct responses can be send to the manager.

bernatgaston:
Flags where not being encoded as expected. Delegation objects are now
being sent to the manager controling the invoke_id used as response and
also the port of the conexion (which was not being stored, and hence,
not allowing multiple managers). It also deals with the race condition
generated by the delegation on the last object sent to the manager
(with no flags) that finishes the flow of cdap objects. To control
this problem, the MA will not sent its last RIB object until all
the delegated objects have finished their work and hence, all the
delegated objects have been send by the MA to the manager. Then,
the MA will send the last response object of the cdap flow for a
given invoke_id
Fixes #932
Fixes #934

Ack @edugrasa @vmaffione

### DIFF
--- a/librina/include/librina/cdap_v2.h
+++ b/librina/include/librina/cdap_v2.h
@@ -222,6 +222,7 @@ class CDAPCallbackInterface
 	virtual void read_request(const cdap_rib::con_handle_t &con,
 				  const cdap_rib::obj_info_t &obj,
 				  const cdap_rib::filt_info_t &filt,
+				  const cdap_rib::flags_t &flags,
 				  const int invoke_id);
 	virtual void cancel_read_request(const cdap_rib::con_handle_t &con,
 					 const cdap_rib::obj_info_t &obj,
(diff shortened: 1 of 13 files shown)


edugrasa:
ACK if you merge the PR I issued you fixing the identation (change seems to introduce no regressions)

edugrasa:
Merging since it is required by the Manager. @vmaffione if you have comment to the PR please state them and Bernat will address them in a separate PR.

vmaffione:
I had a look, it's ok. I didn't have the chance to test, that's why I was waiting to ACK.

edugrasa:
Thanks Vincenzo!

PR #936, opened 2016-05-23
librina: Updated C wrapper

sandervrijders:
The C wrapper now has init and fini functions, to initialize librina
and clean up any flows that weren't deallocated. It also fixes some
other bugs, such as catching all exceptions, as suggested by Marc.


dstaesse:
looks good! Confirmed it works as well...

PR #938, opened 2016-05-25
plugins: lfa: align pft_find method to the null qos-id convention

vmaffione:
When the qos-id in a PFF entry is 0, it means it can serve requests
with any QoS ids. This patch implements this behaviour for the LFA
plugin, similarly to what is done for the PFF default plugin.
Without this patch, the LFA next-hop policy does not work correctly.


sandervrijders:
ACK

PR #940, opened 2016-06-01
Pristine 1.5 fix several memory leaks in robjects

lbergesio:
- internal kobjects were not released due to missing call to decrement ref count
- string name passed to internal kobject functions was not being freed

maintainer: @vmaffione


vmaffione:
Clearly correct --> ACK

PR #941, opened 2016-06-02
kernel:rmt: fix missing clean up of n1port

lbergesio:
add clean up of n1port when the binding operation fails due to wrong RMT's
policy set installed.

This was causing a big kernel dump when trying a second time because a kobject with the same name was already there

maintainer @miqueltarzan


miqueltarzan:
ACK

PR #944, opened 2016-06-02
rinad: ipcm: fixed error in dif allocator

edugrasa:


edugrasa:
Merging since it is a dumb error...

PR #946, opened 2016-06-03
SHIM delegated did not release the lock

### LINKED ISSUE #945: After read with scope it is not possible to create ipcps
A lock appears in the register at diff after a read with scope (and delegation objects) has been done.

bernatgaston:
After a read that target a SHIM, the SHIM was returning an
IPCM_ERROR (it doesn't implement the read) but was not releasing
the lock, which produced a deadlock when trying to registrate
an ipcp to the mentioned SHIM
Fixes #945 
ack @edugrasa

### DIFF
--- a/rinad/src/ipcm/ipcm.cc
+++ b/rinad/src/ipcm/ipcm.cc
@@ -1574,13 +1574,13 @@ ipcm_res_t IPCManager_::delegate_ipcp_ribobj(rina::rib::DelegationObj* obj,
             LOG_ERR("Invalid IPCP id %hu", ipcp_id);
             return IPCM_FAILURE;
         }
+        //Auto release the read lock
+        rina::ReadScopedLock readlock(ipcp->rwlock, false);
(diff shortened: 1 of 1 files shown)


vmaffione:
I don't think you are changing anything with this modification, the bug is probably elsewhere. The lock is released automatically anyway when you return from that function.

edugrasa:
No, the fix works. The lock is acquired by the lookup_ipcp_by_id function, which is later released on scoped lock object destruction. Since the ScopedLock was created *after* checking for the type of IPCP, in the shim case the function left without releasing the ipcp lock.

vmaffione:
Ah I see it, you're right.
Well at least this proves that it is difficult to read :) (I had forgot how it worked)

edugrasa:
Agree, it's difficult to read since the select_ipcp_by_id name doesn't suggest anything about taking locks ;) We've had a number of these errors before already!

edugrasa:
ACK, merging

PR #948, opened 2016-06-06
Added java catch for exceptions thrown in ipc-api

### LINKED ISSUE #947: pristine-1.5: No java exceptions emitted on IPC-API calls in librina
See ipc-api.h:

        /// Reads an SDU from the flow. This function will block until there is an
        /// SDU available.
        ///
        /// @param sdu A buffer to store the SDU data
        /// @param maxBytes The maximum number of bytes to read
        /// @return int The number of bytes read
        /// @throws UnknownFlowException if the port-id is not valid
        /// @throws FlowNotAllocatedException if the flow has been deallocated
        /// @throws InvalidArgumentsException if the arguments of the call are not valid
        /// @throws ReadSDUException if an error happens while reading the SDU
        /// @throws IPCException if an unknown error happens
        int readSDU(int portId, void * sdu, int maxBytes);

No throws in the declaration => no exception code generated by SWIG.
Exception happens => Native code bails with an un-handled exception.

bernatgaston:
If exceptions are not explicetly declared in the .h function
declaration, SWIG does not wrap them and hence, they can not be
catched in java. To avoid the problem without changing the .h,
it is possible to define the catch in the .i.
This commit wraps all the catches thrown by functions in
librina/ipc-api.h
Fixes #947 

ack @edugrasa

### DIFF
--- a/librina/wrap/librina.i
+++ b/librina/wrap/librina.i
@@ -447,6 +447,7 @@ DOWNCAST_IPC_EVENT_CONSUMER(eventWait);
 DOWNCAST_IPC_EVENT_CONSUMER(eventPoll);
 DOWNCAST_IPC_EVENT_CONSUMER(eventTimedWait);
 
+// Allow the callbacks between C++ and java
 %module(directors="1") cdapcallbackjava
 
 %{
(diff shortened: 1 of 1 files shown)


edugrasa:
Ack, merging!

PR #949, opened 2016-06-07
rinad: enrollment: allow the possibility of turning off neighbors enr…

edugrasa:
…oller


edugrasa:
Merging since is a simple change that will be handy for the TNC demo

vmaffione:
Ok but probably you are abusing of this parameter... maybe a new parameter would be better?

edugrasa:
A matter of taste I guess.. if you set it to 0 it won't start, to me it looks clear - otherwise we're making the config file longer - but another parameter may make it even more clear at the cost of making the file longer. Let's leave it like this now and discuss after the demo!

vmaffione:
sure, I don't have a strong opinion, just stimulating the discussion :)
and it is really a minor point

PR #953, opened 2016-06-09
rinad: ma: use proper object name when adding IPCP RIB object into the MA RIB

edugrasa:


edugrasa:
Tested and working, merging since it is needed to allow the Manager to create multiple IPCPs in the same system

PR #955, opened 2016-06-09
rinad: ma: added IPCP enrollment as part of IPCP creation from MA (if requested)

edugrasa:


edugrasa:
Merging, already tested and required for the demo. Now the MAD is feature-complete for the TNC demo.

PR #956, opened 2016-06-11
fix in ipc-api.

dstaesse:
Returning 0 when there is no packet makes the C API think there is a packet with 0 length. When reading in a non-blocking way to loop over port_id's in a select() kind of way, makes the application think there is always a 0-length packet on that port_id.

This should fix ioq3 to work with multiple clients.


edugrasa:
Great :) going to test it tonight!

edugrasa:
Fix, works, merging!

PR #958, opened 2016-06-13
mad: Produce IPCProcess ID

### LINKED ISSUE #952: MA must synchronize with IPCM for delete/create IPCPs
When an IPCP is created/destroyed via the IPCM, the IPCM must update the MA so that it update its RIB.

bernatgaston:
The id of the ipcprocess was being sent by the manager, which is a
wrong behaviour. This characteristic produced also some problems
when erasing the created IPCP etc... Now the original message is a
CREATE(".../IPCProcessID") instead of CREATE(".../IPCProcessID=x").
Now the MAD is able to remove the rib objects when IPCProces dies.

Fixes #952

ack @edugrasa


edugrasa:
CLosing, merged with PR #965

PR #969, opened 2016-06-29
librina: fixed tests compilation

edugrasa:


edugrasa:
Merging since it is just fixing librina tests' compilation failures.

PR #971, opened 2016-06-29
rinad: ipcm: flow_allocation_requested_local: fix DIF selection logic

vmaffione:
The previous logic was wrong, since a DIF specified explicitely from
the user could be overwritten by the DIF allocator.


edugrasa:
Agree, ACK

PR #974, opened 2016-07-11
PFF multi-path policy adaptation to pristine-1.5 

edugrasa:
Adapted find algorithm to support qos value of 0 (wildcard) and cleaned some unused code


edugrasa:
Merging, needed for DC experiments and current version of multipath plugin is not working with pristine-1.5

PR #976, opened 2016-07-13
Pristine 1.5 cbac

edugrasa:
* CBAC policy-set for the security manager
* Updates to required librina and rinad APIs (were not prepared to support access control policies yet, CBAC is the first one)


edugrasa:
ACK, tested scenario explained in the APCC paper, with the config files in https://github.com/ichrak-amdouni/pristine-scenario (slightly updated to match my environment and to the latest version of SDU protection policies). 

Also, verified there are no regressions for the no authentication, SSH2 and TLS authentication plugins.

PR #979, opened 2016-07-17
Pristine 1.5 efcp sysfs

edugrasa:
* Exposed more EFCP variables via sysfs
* Added script to monitor EFCP connections (information exported via sysfs)

Maintainers: @lbergesio


vmaffione:
This seems ok to go in.. any concerns?

PR #981, opened 2016-07-18
Pristine 1.5 dt-utils bug

miqueltarzan:
kernel: dt-utils: Fix a bug in the way the window was re-opened, since it was done outside of the protection provided by the locking mechanism.


edugrasa:
ACK, merging

PR #990, opened 2016-08-29
IPCM console: accept multiple simultaneous connections

jmuchemb:
Writes (responses to clients) are not asynchronous, which means that a client can block the console if it does not read responses. But the buffer is quite big so that's good enough for the moment.


edugrasa:
ACK!

PR #995, opened 2016-09-05
query-rib: filtering by object class/name

jmuchemb:
The netlink message already carried objectClass/objectName fields, but they were neither filled nor processed. 2 things in particular about the implementation:
- If *objectName* does not end with `/`, then *query-rib* returns information for this exact object. Otherwise, it also returns all information from child objects.
- Since the console does not provide any escaping mechanism that could be used to pass an empty token, `-` can be passed to *class* if one only wants to only filter by *name*.

### DIFF
--- a/librina/include/librina/rib_v2.h
+++ b/librina/include/librina/rib_v2.h
@@ -790,13 +790,21 @@ public:
                          const std::string fqdn);
 
         ///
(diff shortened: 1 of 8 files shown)


edugrasa:
Nice one, ack, tested!

PR #997, opened 2016-09-06
kernel: shim-tcp-udp: quick'n dirty way to update directory entries

jmuchemb:
This is waiting for the new shim-udp that will support multiple flows and do DNS queries.


edugrasa:
Hi Julien,

Before merging this one, could you describe an example of how the directory entries need to be configured after this change? Thanks!

jmuchemb:
Done: I amended the commit message with examples.

edugrasa:
Finally had the time to test this, works well. Nice job Julien!

PR #1003, opened 2016-09-09
rina-echo-time: fix processing of late pongs

jmuchemb:
- Timeouts were reported as 'bad response'.
- Late pongs (i.e. not lost, which can happen when testing over shim tcp)
  were also reported as 'bad response', except for a few specific SDU sizes.

There remains 1 issue that I explained in a comment:
> TODO: Always read SDUs while waiting before the next ping. Otherwise, any late and bad response would flush 'sent' and we'd have no way anymore to check correctly the following responses. This is a rare scenario and for now, readSDU can't do that without using all CPU, which we want to avoid.

Do you know if `IPCManager::readSDU` returns `-EINTR` when a signal is processed ? In `Client::readSDU`, I could use replace the busy loop by something using `setitimer`.

The following output is a test over shim tcp, and firewall rules to temporarily drop packets (I used `ts -s %.S` to prefix with timestamps):
```
00.003897 SDU size = 20, seq = 0, RTT = 52.741 ms
01.001654 SDU size = 20, seq = 1, RTT = 57.667 ms
02.058324 SDU size = 20, seq = 2, RTT = 56.036 ms
03.118309 SDU size = 20, seq = 3, RTT = 59.413 ms
06.121567 8972(1473415029)#rina-echo-time (WARN): Timeout waiting for reply, SDU maybe lost
08.128222 8972(1473415031)#rina-echo-time (WARN): Timeout waiting for reply, SDU maybe lost
10.131558 8972(1473415033)#rina-echo-time (WARN): Timeout waiting for reply, SDU maybe lost
12.134310 8972(1473415035)#rina-echo-time (WARN): Timeout waiting for reply, SDU maybe lost
14.141571 8972(1473415037)#rina-echo-time (WARN): Timeout waiting for reply, SDU maybe lost
16.148235 8972(1473415039)#rina-echo-time (WARN): Timeout waiting for reply, SDU maybe lost
18.149500 8972(1473415041)#rina-echo-time (WARN): Timeout waiting for reply, SDU maybe lost
18.760593 SDU size = 20, seq = 4, RTT = 14642 ms
18.760818 SDU size = 20, seq = 5, RTT = 12639 ms
18.761309 SDU size = 20, seq = 6, RTT = 10633 ms
18.761893 SDU size = 20, seq = 7, RTT = 8630.1 ms
18.762450 SDU size = 20, seq = 8, RTT = 6627.9 ms
18.762907 SDU size = 20, seq = 9, RTT = 4621.1 ms
18.763346 SDU size = 20, seq = 10, RTT = 2614.9 ms
18.763745 SDU size = 20, seq = 11, RTT = 609.23 ms
19.815001 SDU size = 20, seq = 12, RTT = 50.495 ms
20.874966 SDU size = 20, seq = 13, RTT = 59.303 ms
```

### DIFF
--- a/rina-tools/src/rina-echo-time/et-client.cc
+++ b/rina-tools/src/rina-echo-time/et-client.cc
@@ -28,6 +28,7 @@
 
 #include <cstring>
 #include <cmath>
+#include <deque>
 #include <iostream>
 #include <sstream>
 #include <cassert>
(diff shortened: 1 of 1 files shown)


edugrasa:
ACK

PR #1006, opened 2016-09-14
Pristine 1.5 pff: fix locking and atomic modify

edugrasa:
Fixed locking in PFF policy implementations (default and multipath), as it is right now there can be soft lockups if the RMT queries pff->nhop while the pff is being updated.

Also added a pff_modify atomic operation, that updates all the entries in the PFF without releasing the PFF lock. As it is done right now, first all the PFF entries are removed, then lock released, then for each entry get lock, add entry and release lock. If a PDU queries the next hop in any intermediary step (getting the lock), it may be dropped because the PFF entry is still not there.

Maintainers: @sandervrijders, @vmaffione

### DIFF
--- a/linux/net/rina/ipcp-instances.h
+++ b/linux/net/rina/ipcp-instances.h
@@ -324,6 +324,9 @@ struct ipcp_instance_ops {
 
         int (* pff_flush)(struct ipcp_instance_data * data);
 
+        int (* pff_modify)(struct ipcp_instance_data * data,
+                           struct list_head * entries);
+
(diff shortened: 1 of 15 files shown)


vmaffione:
I think we should change the PFF LFA, shouldn't we?

edugrasa:
Ups forgot about this one :) Will update the pull request tomorrow morning

edugrasa:
Updated!

vmaffione:
my regression tests are ok --> ACK

PR #1012, opened 2016-09-18
rina-echo-time: remove memory leak

vmaffione:
The operation removed is not needed, because it is performed anyway by the Client constructor.
It also causes a memory leak, since nobody frees tmp.

@miqueltarzan @edugrasa


edugrasa:
ACK!

dstaesse:
Can you remove the brackets on lines 77-78 as well?

vmaffione:
too late :)

Issue #998, opened 2016-09-06
POSIX API for applications


edugrasa:
Redesign RINA API for applications so that it is closer to the POSIX standard.

edugrasa:
First version already completed

PR #1013, opened 2016-09-19
rina-echo-time: do not use all CPU while reading a SDU with a timeout

jmuchemb:
This implements what I suggested in #1003, waiting that SDU can be read/written with posix API (like select).
This was quick enough to do, and the use of all CPU annoyed me too much, in particular in VMs.

### DIFF
--- a/rina-tools/src/rina-echo-time/et-client.cc
+++ b/rina-tools/src/rina-echo-time/et-client.cc
@@ -26,6 +26,7 @@
 // SUCH DAMAGE.
 //
 
+#include <csignal>
 #include <cstring>
 #include <cmath>
 #include <deque>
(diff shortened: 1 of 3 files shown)


rina-tools/src/rina-echo-time/et-client.cc
```diff
@@ -475,25 +513,15 @@ int Client::readSDU(int portId, void * sdu, int maxBytes, unsigned int timeout)
...
-			throw e;
-		}
-	}
+	ReadSDUWithTimeout _(portId, timeout);
+	do {
+		int bytes_read = ipcManager->readSDU(portId, sdu, maxBytes);
+		if (bytes_read != -EINTR)
+			return bytes_read;
+		get_current_time(endtp);
+	} while ((unsigned) time_difference_in_ms(begintp, endtp) < timeout);
```
dstaesse: this looks dangerous... better make the timeout signed!

dstaesse:
This should probably be fixed in the kernel with a waitqueue, not in the application.  Setting a timeout value could be added in the flow_io_ctl syscall, for instance.

jmuchemb:
This is better because it would be easier for other applications to do the same thing.
However, that remains a hack:
- the timeout value often changes all the time, and a separate syscall just for that is bad for performance
- also, a timeout often has nothing to do with a flow, but it's rather a way to have timers in parallel to IO operations
- is it worth doing this waiting for #998 ? (in any case, I won't do it)

rina-tools/src/rina-echo-time/et-client.cc
jmuchemb: do you prefer `(long) time_difference_in_ms(begintp, endtp) < (long) timeout` ?

dstaesse: why not just ```while (time_difference_in_ms(begintp, endtp) < (unsigned) timeout);```?

dstaesse:
Maybe I misunderstood what you wanted to do. I thought you needed a blocking read with a timeout, like setsockopt SO_RCVTIMEO does. I don't really see performance issues, unless you change the timeout for every read, but I wouldn't know why you need that.

rina-tools/src/rina-echo-time/et-client.cc
jmuchemb: The type of `timeout` is already `unsigned`, so I guess you mean `_difference_in_ms(begintp, endtp) < timeout`.
I don't know why there was cast to an integral type; my commit does not change the condition (well, off by 1ms). A more important issue is the use of `get_current_time` (i.e. `gettimeofday`), whereas we'd prefer to use a monotonic clock for timeouts.

jmuchemb:
What you suggest is indeed equivalent to SO_RCVTIMEO, but I never program with this socket option. I find `select` and friends more versatile and enough.

> you change the timeout for every read, but I wouldn't know why you need that.

Most IO loops do that all the time. A few examples:
- NEO: https://lab.nexedi.com/nexedi/neoppod/blob/c39d5c670e55f4f739d4dd633fe862c37ecde9ab/neo/lib/event.py#L166 (the computed timeout is passed to `self.epoll.poll` below)
- re6stnet: https://lab.nexedi.com/nexedi/re6stnet/blob/674cad325677d9ae411a8bc4bcd8ea1222c26bc7/re6st/utils.py#L176 (1 line to compute the timeout)
- babeld: https://github.com/jech/babeld/blob/babeld-1.7.1/babeld.c#L589 (the select is at line 626)

And in fact, in rina-echo-time, I should also do the same between pings (i.e. when read_late is true). Instead of always passing `wait` as timeout, I should ideally:
1. compute next_ping = now + wait
2. repeat readSDU(timeout=next_ping-wait) until it returns -EAGAIN

I just found that this PR is somewhat related to c469b0e86f5d881a01626b76536bbf42e48813a2

dstaesse:
ok, fine, let's merge this in

rina-tools/src/rina-echo-time/et-client.cc
dstaesse: sorry, my bad, should be signed (see my first response)
while (time_difference_in_ms(begintp, endtp) < (long) timeout)

PR #1028, opened 2016-10-17
Merge C++ verison of CDAP connect

mcrotty:
Should work for creating IPCPs


edugrasa:
Has to be issued against pristine-1.5, not master

PR #1029, opened 2016-10-17
Merge C++ verison of CDAP connect

mcrotty:
Should work for creating IPCPs.


edugrasa:
ACK, verified functionality in small scale (two systems with a MA each). Merging and testing in larger experimental setup via the demonstrator.

PR #1031, opened 2016-10-19
Pristine 1.5 cc: fix bug and use logging

edugrasa:


edugrasa:
Merging, fixing bug in experimentation

PR #1032, opened 2016-10-21
librina: added missing include; rinad: added thread names

edugrasa:


edugrasa:
Merging to fix build problem with buildroot

lbergesio:
Sadly, this did not solve the problem for me :(

Issue #1038, opened 2016-11-16
Refactor IPCP Daemon to facilitate shim IPCP daemons in user space


edugrasa:
This way shims can easily have also user-space components (requirement for the shim DIF over WiFi)

edugrasa:
Fixed by #1039

PR #1039, opened 2016-11-16
Arcfire ipcp factories

edugrasa:
Addresses #1038 

@vmaffione, @lbergesio could you take a look?

### DIFF
--- a/librina/include/librina/enrollment.h
+++ b/librina/include/librina/enrollment.h
@@ -105,7 +105,7 @@ public:
 
 	/// Process a request to initiate enrollment with a new Neighbor, triggered by the IPC Manager
 	/// @param event
-	virtual void processEnrollmentRequestEvent(EnrollToDAFRequestEvent * event) = 0;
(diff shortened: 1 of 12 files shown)


vmaffione:
I had a review, you've moved the code in a separate module, adding the base class, that's ok. I did regression tests and those are ok. What does @lbergesio think?

lbergesio:
I think it is ok for the time being. Withi this abs class we can already create a different ipcp implementing probably the enrollment task equivalent for the shim wifi. In the future I may go a bit further, gather normal-ipcp code in a dedicated directory and maybe provide abstract classes for the ipcp components too. But for this, we probably need to end 100% pristine-1.5 development and have better knowledge of what it will be required for, so far, by the shim-wifi.

edugrasa:
Agree Leo, this is a first step that allows the development to start. Once we know more, pristine-1.5 is closed and the renumber code is merged in we can go further. Merging!

PR #1040, opened 2016-11-16
rinad: updated DIF allocator to handle same app registered in many DIFs

edugrasa:
And automatically pick the right one. Tested, seems to work ok with no regressions.

@vmaffione could you take a look? Thanks!


vmaffione:
It looks ok, I will also test it.

vmaffione:
My tests are OK, can you confirm it works in the VPN scenario we discussed?

vmaffione:
Anyway, we can merge in the meanwhile

edugrasa:
Yeap, tomorrow I will test it!

2016-11-16 22:40 GMT+01:00 Vincenzo Maffione <notifications@github.com>:

PR #1045, opened 2016-11-25
rinad:Move KernelSyncTrigger to ipc-process.h

lbergesio:
class KernelSyncTrigger moved to ipc-process.h and modified to accept an
AbstractIPCProcessImpl* so it can be used by different implementations of IPCP
daemons.


edugrasa:
ACK!

PR #1046, opened 2016-11-25
Pristine 1.5 kmgmt

edugrasa:
Code for Key Management demo: includes key management application (with Central Key Manager and Key Management Agents). Added minor updates and fixes to librina in order to allow applications other than IPCPs to properly use the authentication policies and carry out SDU protection in user space.


edugrasa:
Merging to pristine-1.5 for the Key Mgmgt demo

PR #1055, opened 2017-01-17
Add receive-side poll() kernel support

vmaffione:
This patch adds the necessary mechanism to implement selective wake up on poll()/select()/epoll() userspace callers with POLLIN events.
When a packet is received on the flow (or the flow is deallocated), the poll()er is waken up so that it can react to the event.

This depends on (and includes) #1054


edugrasa:
Tested! Works as expected and no regressions

PR #1056, opened 2017-01-19
librina: core: fix race condition on eventQueueReady

vmaffione:
The race condition may happen since the thread that pops the event
tries to read from the eventfd before the thread that pushed the
event had the chance to write to the eventfd.


edugrasa:
Fixes the issue!

PR #1058, opened 2017-01-19
librina: core: use a better scheme for eventfd notifications

vmaffione:
The last version of eventfd management was race-condition free, but affected by possible
spurious wake-ups, since the notifications (eventfd write) used to
happen before the event was pushed.
This new solution manages the race condition between write() and
read() by simply allowing the read() to block.

@edugrasa can you please verify that the race condition is not there and that there are no regressions?


edugrasa:
Tested with multiple applications, seems to work fine with no regressions

PR #1059, opened 2017-01-20
kernel: qta_mux: updated to new kernel apis

edugrasa:


edugrasa:
Merging to allow qta_mux plugin to compile

PR #1071, opened 2017-02-23
Allow an IPCP to enroll to a DIF without having to provide information on a specific DIF member

### LINKED ISSUE #1070: Allow enrollment to a DIF without specifying specific IPC Process name / instance
"Full implementation" requires N-1 DIFs to support whatevercast (can be a problem for shims..). 

An easier first step is to update the registration API so that applications can pass its appl. name and optionally the DAF name they belong to when they register to a DIF. The DIF will create two entries in the Directory Forwarding Table (one for the DAF name and one for the specific app name). This capability will only be added to IPCPs for now.

Enrollment will also be updated to be able to handle enrollments to a specific member of the DAF (as it is now) or to any member of the DAF (to the DAF name).

This is required for the mobility work, since when IPC Processes in mobile hosts join DIFs at new access points, they will not know the name of the specific IPCP they are attaching two until the application connection with that IPCP has completed.

edugrasa:
Allow an IPCP to enroll to a DIF without having to specify a specific IPCP member, fix #1070. To do so, when an IPCP registers to an N-1 DIF two entries are added in the N-1 DIF directory forwarding table, one with the IPCP name/instance, and one with the N-DIF name. 

Works for DIFs over normal DIFs and for DIFs over the shim DIF for Ethernet (can be added later for the shim DIF over TCP and UDP and shim DIF for HV later on if needed).

To test try the enrollment command from the IPCM console without specifying the IPCP name /instance, for example:

````
enroll-to-dif 3 normal.DIF 310
````

Enrollment with the usual syntax (to a specific DIF member) keeps working as before; for example:

````
enroll-to-dif 3 normal.DIF 310 test2.IRATI 1
````


vmaffione:
Regression tests failed (default demo.conf).

On node A I run `rinaperf -l`, on node C I run `rinaperf`. What I get on node A:
````
395(1489251373)#ipcp[2].rib-daemon (DBG): Got message of 517 bytes, handling to CDAP Provider
395(1489251373)#ipcp[2].rib-daemon (DBG): Received A-Data CDAP message from address 17 
Opcode: 4_M_CREATE
Flags: 0
Invoke id: 2
Object class: Flow
Object name: /fa/flows/key=17-3
Scope: 0

395(1489251373)#rib (DBG): Found a callback for create operation of class 'Flow' and path '/fa/flows'
395(1489251373)#ipcp[2].flow-allocator (ERR): The directory forwarding table returned no entries when looking up rinaperf-data:server:::
395(1489251373)#librina.syscalls (DBG): Invoking SYS_readManagementSDU (329)
````

edugrasa:
Should be fixed now, can you try again? Thanks!

vmaffione:
Ok, no regressions.
The broadcast enrollment also works (removing the last argument from enroll.py in the demonstrator repo).

PR #1073, opened 2017-03-08
Kernel upgrade from 4.1.33 to 4.1.38

vmaffione:
Periodic upgrade, we need to check that there is no regression.


vmaffione:
My regression tests are ok.

edugrasa:
Mine too, merging!

PR #1074, opened 2017-03-08
rina-tools: use large read() in rina-echo-time client

vmaffione:
Use a large buffer for read().
This prevents error related to partial SDU read, which happens if packets larger than expected come on the flow.

@edugrasa


edugrasa:
ACK!

PR #1083, opened 2017-03-15
irati-ctl: Add --version argument

lbergesio:
Add --version argument to irati-ctl to avoid the make check performed by the
install-user-from-scratch script to fail.

It has to be implemented by defining a custom Action since the builtin version
option prints the message in stderr and is taken as error by the test.

### DIFF
--- a/rina-tools/src/irati-ctl
+++ b/rina-tools/src/irati-ctl
@@ -9,6 +9,8 @@ import socket
 import time
 import re
 
+VERSION="1.0.0"
+
 def printalo(byt):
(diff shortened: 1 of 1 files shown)


lbergesio:
@vmaffione, sure, let's go with yours, as I said the simplest solution was using the 'version' action but the problem was that with it the text was sent to stederr and considered failed by the test, so I came with this, I agree, cumbersome thing.

vmaffione:
I understand, no problems! :)
Thanks

PR #1086, opened 2017-04-06
Arcfire wifi: first release

edugrasa:
Includes:

* Several bug fixes to enrollment, CDAP and RIB Daemon
* Support for shim wifi IPCP interfacing with the WPA Supplicant
* A draft mobility manager that managers handovers in a hardcoded scenario 

@vmaffione could you check that it introduces no regressions? Testing with wifi requires a lengthy setup and we've already done an extensive number of tests


vmaffione:
Checked, I don't see regression.
Anyway, how would one be expected to test this?

edugrasa:
To test the evaluation scenario you need three laptops, one with two WiFi interfaces (the "mobile host"), two more with one (access points configured with hostApd). One AP manages a WiFi network with SSID "arcfire", the other one another WiFi network with SSID "pristine". You'd need the host apd as well as WPA supplicant conf files to test as well. 

We'll document it all for the May demo in an ARCFIRE wiki page (so that partners can easily replicate it as well), and then make it public as part of the experiments.

vmaffione:
Ok, thanks for the details. In the meanwhile, let's merge.
We should definitely do the demo in a way that the node goes from pristine --> arcfire !

PR #1088, opened 2017-04-22
gpb: specify syntax=proto2 in all files to avoid warnings on newest GPB compilers

vmaffione:
@edugrasa can you give a try?


edugrasa:
ACK, tested!

PR #1091, opened 2017-04-25
rina-tools: rinaperf: large refactoring towards automated tests

vmaffione:
With this patch rinaperf client is now able to recover server-side
measurements, so that it can print complete results at the end of
the test, similarly to what netperf does.

This is a code alignment from rlite.

To check the new rinaperf cababilities, try "unidirectional forwarding tests"
````
    $ rinaperf -t perf -s 1000 -c 10000
````
or try "request/response test"
````
   $ rinaperf -t rr -s 200 -c 10000
````

Added full support for duration tests (e.g. 4 seconds test):
````
   $ rinaperf -t rr -s 200 -D 4
````

It is possible to use Ctrl-C any time to stop the tests. This will stop the clients and still perform handshake with the server to receive results on the control connection.

### CODE REVIEW

File: rina-tools/src/rlite/rinaperf.c [hunk 1]
```diff
@@ -560,24 +791,160 @@ static void *
...
+            printf("Configuring test type %u, SDU count %lu, SDU size %u, "
+                    "ticket %u\n",
+                   cfg.opcode, cfg.cnt, cfg.size, ticket);
+        }
+
+        /* Wait for the client to allocate a data flow and come back to us. */
+        pthread_mutex_lock(&rp->ticket_lock);
+        clock_gettime(CLOCK_REALTIME, &to);
+        to.tv_sec += 5;
+        ret = pthread_cond_timedwait(&w->data_flow_ready,
```
sandervrijders: Doesn't take spurious wakeups into account.

vmaffione: The code itself cannot produce spurious wakeups. Why should those happen?

rina-tools/src/rlite/rinaperf.c
```diff
@@ -40,179 +40,172 @@
...
 struct worker;
 
-typedef int (*perf_function_t)(struct worker *);
+struct rp_config_msg {
+    uint32_t opcode;    /* opcode: ping, perf, rr ... */
+    uint32_t ticket;    /* valid with RP_OPCODE_DATAFLOW */
+    uint32_t size;      /* packet size in bytes */
+    uint64_t cnt;       /* packet/transaction count for the test
+                         * (0 means infinite) */
+};
```
sandervrijders: Should these take the packed attribute?

vmaffione: yes, let me update the code to the latest rlite version

vmaffione: Yes, imported this from upstream

rina-tools/src/rlite/rinaperf.c
sandervrijders: It's because of the threading implementation in the kernel. Even if you didn't call pthread_cond_signal, it can still wake up. So you need to do a while(cond == false) cond_wait();

vmaffione: Right, but new code uses a semaphore, so the problem goes away

vmaffione: Right, but new code uses a semaphore, so the problem goes away

sandervrijders:
Code looks okay for my part now. Someone should probably give it a testdrive before merging in. Nice work.

edugrasa:
@vmaffione Is this the expected output?

````
root@e2-edu-vm2:/usr/local/irati/bin# ./rinaperf -t perf -s 1000 -c 10000
Starting unidirectional throughput test; message size: 1000, number of messages: 10000, duration: inf
                Packets       Kpps       Mbps
Sender            10000 -92559641157289301412905710012271939667257667601819249288413184.000 -92559653468829427386383975183145674369325473325413350976258048.000
Receiver           9997 -867922223097221224170017113291268788404647866235209870625586576885639852214536131322856918119494405640636881214763330781572396345868635964755114433807692685370448134861266939312105851240497679642384340237680640.000 -867922380449640072109588233917211357738677018993295598101786998264009002201737495477253744996893415289782686515456160647627356059331420560947798366363476034531869462960682269729809047908283112609995950456307712.000
root@e2-edu-vm2:/usr/local/irati/bin# 
````

vmaffione:
well, no of course. In which testbed/scenario did you produce that?
What does it happen if you do a "rr" test rather than a "perf" test?

vmaffione:
@edugrasa Btw, I also did an update, can you retry?

edugrasa:
Same results, I'm just trying rinaperf in  a Normal over shim-eth-vlan scenario, just two IPCPs directly connected.. May I do something wrong? 
I'm running it on XEN VMs.. (just to give more details of my setup)

edugrasa:
More output:

`````root@e2-edu-vm2:/usr/local/irati/bin# ./rinaperf -t rr -s 200 -D 4
Starting request-response test; message size: 200, number of messages: inf, duration: 4 secs
              Transactions       Kpps       Mbps    Latency (ns)
Sender               16181      0.000      0.000               0
root@e2-edu-vm2:/usr/local/irati/bin# 
````

vmaffione:
I'll retry with the demonstrator. YOu can at least add "-v" option to server to see if there is more useful information. Have you tried the updated code?

edugrasa:
with -v : 

````
root@e2-edu-vm2:/usr/local/irati/bin# ./rinaperf -t rr -s 200 -D 4
Starting request-response test; message size: 200, number of messages: inf, duration: 4 secs
              Transactions       Kpps       Mbps    Latency (ns)
Sender               16510      0.000      0.000               0
root@e2-edu-vm2:/usr/local/irati/bin# 
````
* Server:
````
root@e2-edu-vm3:/usr/local/irati/bin# ./rinaperf -l -v
Configuring test type 1, SDU count 0, SDU size 0, ticket 200
Stopped remotely
received 16511 PDUs out of 0
````

vmaffione:
mmm, there is something wrong,the value for SDU size is corrupted, and swapped with the ticket.
In the server log, ticket should be 0, and SDU size 200.

vmaffione:
I cannot reproduce this issue on my QEMU IRATI VMs, it works just fine to me...

vmaffione:
Are the VMs 32 bits?

edugrasa:
no 64...

edugrasa:
could you check again that it works in the demonstrator with the current changes? if so I'llmerge, I guess I may have messed up something in my machines,..

vmaffione:
I've checked again, both with buildroot freshly-generated images and legacy mode images. This works ok.
In addition to IRATI, it also works on rlite, on both demonstrator and physical machines (and jfed).

@sandervrijders suggestion was that it may be related to the VM clock in your Xen environment...
But problem is that I cannot reproduce it. You may want to debug a bit to check that time differences done by rinaperf are ok (e.g. reasonable numbers)

PR #1093, opened 2017-05-11
Arcfire cdaprel

edugrasa:
Changes:

* Use of optional reliable N-flow for N-IPCP layer management task (this feature is off by default, can be enabled per DIF via the configuration of the EnrollmentTask PS)
* Fixed bugs in the shim-eth-vlan
* Reduced overhead of NL message parsing (in terms of code)
* Fixed bugs in enrollment task
* Updated mobility manager to OMEC demo scenario
* Fixed bug when using reliable flows and calling dev_queue_xmit

@vmaffione could you test for regressions? Thanks :)


vmaffione:
no regressions

PR #1098, opened 2017-05-29
IP over RINA flows

edugrasa:
Enable allocation of IP over RINA flows via the IPCM console. The flow is exported as a point-to-point network device and made available to the IP layer via configuration of the IP routing table. The network admin must register one or more IP prefixes via the IPCM to act as targets of flow requests.

Example use case: two systems, each one with two Ethernet interfaces. One interface is allocated a VLAN and will be used by RINA (via the shim-eth-vlan), the other interface is assigned an IP address. For example, one ethernet interface at system A is assigned IP address 10.10.90.1 (netmask 255.255.255.0), and another ethernet interface at system B is assigned IP address 10.10.100.1. Both systems are physically connected via the Ethernet interfaces that have the VLAN assigned. After initiating RINA in a usual normal over shim Ethernet configuration, the steps to allocate an IP flow in this example are (assuming enrollment is already in place):

* Log in the IPCM console at system B, register the IP prefix at the DIF 'normal.DIF': register-ip-prefix 10.10.100.0 24 normal.DIF
* Log in the IPCm console at system A, request the allocation of the IP over RINA flow between prefixes 10.10.90.0/24 and 10.10.100.0/24 using the DIF 'normal.DIF': allocate-iporina-flow 10.10.90.0 24 10.10.100.0 24 normal.DIF

Now you can ping 10.10.100.1 from system A and 10.10.90.1 from system B. If you type 'ifconfig' you will see the RINA device, named rina<ipcp_id>.<port_id>, and if you type 'ip route show' you sill see the route to the address prefix going through the RINA device. To deallocate the flow, just login to any of the two IPCM consoles and type:

* deallocate-iporina-flow <port_id of the flow>

Routing table entries will be removed and the RINA device(s) at both ends unregistered. 

@vmaffione, could you test it? Thanks!


edugrasa:
@vmaffione could you merge as is? Will move the management of "IP RINA apps" out of the IPCM into a dedicated application later on, now we have other more urgent work and cannot allocate time to it. It is only a first version anyway, needs to be enhanced to support multiple VPNs by leveraging network namespaces...

vmaffione:
Regression tests ok.

PR #1110, opened 2017-07-25
Arcfire nonetlink

edugrasa:
- Removed Netlink infrastructure, now IRATI uses a control device for the user-space kernel and process to process communication. 

- Followed the serdes approach of rlite, sharing code between user-space and kernel. 

- Reimplemented the rina-api to fix the bugs on parallel flow allocation

- Rearranged some workflows between apps and IPC manager.

- Added the rina-cat application, implemented by Steve Bunch

@vmaffione could you test eventually? No rush. I've checked with a number of demonstrator examples (including the isp security one), all features seem to be working well. Also, the bug with the parallel flows it's gone in this version.


vmaffione:
Just tested the demonstrator in legacy mode.
It works way better than before, that is now rinaperf application is responsive to Ctrl-C (similar to rlite).
Well done!

PR #1112, opened 2017-07-27
Arcfire staticrouting

### LINKED ISSUE #1111: Add routing policy to allow static routes and commands to the CLI to add or remove these routes
The rationale being that static routes are a very interesting option for some scenarios.

edugrasa:
Fix #1111 . Adds a static routing policy. Routing entries are specified in the config file, as parameters of the routing policy. Example:

````
     "routingConfiguration" : {
         "policySet" : {
           "name" : "static",
           "version" : "1",
           "parameters" : [{
             "name"  : "17",
             "value" : "0-1-17"
           }]
     }
````
Each policy parameter is a routing table entry. The name of the parameter is the destination address, the value of the parameter has the format <qos_id>-<cost>-<next hop address>.

Installing routing and forwarding table entries can be queried from the IPCM console, using the "query-rib" command.

````
Name: /resalloc/nhopt; Class: NextHopTable; Instance: 11
Value: -

Name: /resalloc/nhopt/key=17-0-1-17; Class: NextHopTableEntry; Instance: 37
Value: Destination name: ; Addresses: 17; ; QoS-id: 0; Cost: 1; Next hop addresses:  17; / 

Name: /resalloc/pduft; Class: PDUForwardingTable; Instance: 12
Value: -

Name: /resalloc/pduft/key=17-0-6; Class: PDUForwardingTableEntry; Instance: 38
Value: Destination address: 17; QoS-id: 0; Port-ids to be forwarded: 6/ 
````
@miqueltarzan could you test to make sure it complies with your experimentation requirements? Thanks!

### DIFF
--- a/librina/src/ipc-process.cc
+++ b/librina/src/ipc-process.cc
@@ -1032,7 +1032,13 @@ RoutingTableEntry::RoutingTableEntry(){
 const std::string RoutingTableEntry::getKey() const
 {
 	std::stringstream ss;
-	ss << destination.name << "-" << qosId << "-" << nextHopNames.front().alts.front().name;
+	if (destination.name != "") {
+		ss << destination.name << "-" << qosId << "-" << cost << "-"
+		   << nextHopNames.front().alts.front().name;
+	} else {
+		ss << destination.addresses.front() << "-" << qosId << "-" << cost << "-"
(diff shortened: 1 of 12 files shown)


miqueltarzan:
Tested and works. ACK!

PR #1120, opened 2017-08-31
Add alternate Syntax for shim-tcp-udp dirEntry and expReg configuration

TRIA:
Alternate syntax for shim-tcp-udp dirEntry and expReg config entries  …
Add parsing of an additional syntax for config entries.  It is no
longer necessary to count characters in the fields.  To smooth the
transition to the new syntax, the old syntax is still accepted.

In the new syntax, every field begins with ':' including the initial
count of the number of entries.  This distinguishes new-syntax
config strings from old ones, as old syntax configuration strings begin with
a number and new syntax ones with a colon.  The syntax in use for a string is
automatically detected by examining the first character.  You can use
a different syntax on dirEntry and expReg, they are independent.
Old syntax:
	#entries:count:fieldcount:fieldcount:fieldcount:field
  vs. new syntax:
	:#entries:field:field:field:field

Example:
old:	
    "dirEntry" : "1:9:razzycam40:13:192.168.1.1154:2426",  
    "expReg" : "1:9:razzycam20:4:2426"

new:
    "dirEntry" : ":1:razzycam4::192.168.1.115:2426",  
    "expReg" : ":1:razzycam2::2426"

The meaning of "#entries" above, which is the number of configuration
entries in the string, is unchanged.  They are simply concatenated (the
#entries field is not repeated). 
Example:
	"dirEntry" : ":2:razzycam4::192.168.1.115:2426:razzyr::192.168.1.149:2426",

The colon is an illegal character in any field in the new syntax, but no other
differences in the resulting kernel config data result from this change.  If this
presents a problem, the configuration string must use the old syntax.

Other changes:
Moved common code in parse_assign_conf() into a new common function
get_syntax_and_count() that also detects the new vs. old syntax
and takes care of freeing a kernel copy in event of an error.

Fixed latent bugs in detection of illegal entry count.

Added additional INFO logfile output to aid in debugging configurations.


edugrasa:
Tested and works well! Regarding the bug of the shim-tcp-udp, it seems it corrupts SDU sizes smaller than 460 bytes, but if thay are bigger than that it works well. I'll investigate more at the end of this week. 
MErging this one!

PR #1123, opened 2017-09-06
Fix Makefile.in to get RTX back to work

miqueltarzan:
There were missing variables in the Makefile.in file that prevented the loading of critical functions for the retransmission of PDUs. 

It's been tested with rina-echo-time. There are no foreseable problems with other applications as the changes affect only EFCP's hooks.  

@edugrasa please, could you review and merge?


edugrasa:
Tests are ok bug is gone and no regressions!

PR #1126, opened 2017-09-19
Support for 4.1 - 4.9 LTS kernels, build system update

edugrasa:
* Added support for Linux kernel versions 4.1 to 4.9
* Updated build system to use configure and make, removed old scripts

@miguelpdl could you give it a try on the Ubuntu systems that were failing before? README has been updated with install instructions for the new build system, basically just:
````
./configure --prefix <path to IRATI installation> 
make install
````
The --prefix argument is optional.

### CODE REVIEW

**Comment by miguelpdl** [comment]
No successful build yet, the configure script pops out at the `rinad phase` with

```
checking for LIBRINA... no
configure: error: Your system lacks of librina support
Cannot complete rinad phase
```

When I look to the configure script created under rinad, the line in the configuration causing the issue is 

```
fi
        if test $_pkg_short_errors_supported = yes; then
                LIBRINA_PKG_ERRORS=`$PKG_CONFIG --short-errors --print-errors --cflags --libs "librina >= 0.2.0" 2>&1`
        else
                LIBRINA_PKG_ERRORS=`$PKG_CONFIG --print-errors --cflags --libs "librina >= 0.2.0" 2>&1`
        fi
        # Put the nasty error message in config.log where it belongs
        echo "$LIBRINA_PKG_ERRORS" >&5


    as_fn_error $? "Your system lacks of librina support" "$LINENO" 5
```

**Comment by miguelpdl** [comment]
Thanks for the change, we've got further along now but there's is a stop of the build at

```
/bin/bash ../libtool   --mode=install /usr/bin/install -c   librinajsoncpp.la librinairati.la librina.la '//lib'
libtool: install: /usr/bin/install -c .libs/librinajsoncpp.so.0.0.0 //lib/librinajsoncpp.so.0.0.0
/usr/bin/install: cannot create regular file '//lib/librinajsoncpp.so.0.0.0': Permission denied
```
and then 

> make: *** [install] Error 2
> Cannot complete librina phase

However as discussed this is due to the fact that LIBRINA needs to be in place for RINAD to be build and therefore the library LIBRINA has to be installed. This should be fixed up.

In the mean time I will run this under sudo

> sudo ./configure

**Comment by edugrasa** [comment]
Yes, it's a step in the good direction (so that you can build all using configure and make install), but not the final solution yet (have no time for that right now). rinad and rina-tools configure script expect librina to be installed on the system.. Has to be fixed in the future.

**Comment by edugrasa** [comment]
Merging, finally works

**Comment by miguelpdl** [comment]
On a Debian 9 machine, had a build fail with

```
irati/stack/librina/src/security-manager.cc:930:15: error: invalid use of incomplete type ‘DH {aka struct dh_st}’
```

This turns out to be an issue with the default libssl-dev installed in Debian 9. Having fixed that, other issues arose the main one being libprotoc which is version 10 on Debian 9, but the IRATI stack  is only compatible with libprotoc9_2.6.1.

CAn confirm that this will NOT build on Debian 9

**Comment by miguelpdl** [comment]
Did a build on Ubuntu 16.04. Most things seems to run fine until an error at the end

```
INSTALL /home/kodacall/src/irati/stack/plugins/red/red-plugin.ko
At main.c:222:
- SSL error:02001002:system library:fopen:No such file or directory: bss_file.c:175
- SSL error:2006D080:BIO routines:BIO_new_file:no such file: bss_file.c:178
sign-file: certs/signing_key.pem: No such file or directory
  DEPMOD  4.4.0-93-generic

```

In looking at this issue it appears to be caused by a new and stricter method of kernel module signing on Ubuntu 16.04. More can be read here about [Ubuntu 16.04, make install leads to ssl error](https://github.com/free5lot/hid-apple-patched/issues/23).

Does this mean a new issue should be raised against the IRATI Stack, in regards to DKMS installation being the recommended way on Ubuntu ?

PR #1128, opened 2017-09-24
Reliable flows fixes

edugrasa:
* Fix bug in reliable flow destruction
* Fix estimation of RTT
* Fix timing of retx, implement exponential backoff


edugrasa:
Merging, fixes important bugs in rtx control

PR #1139, opened 2017-10-05
Arcfire, link-state routing bug fix

edugrasa:
Fix bug in routing policy (link state)


edugrasa:
Tested in RPi testbed

PR #1158, opened 2017-10-25
1109 travis ci build

miguelpdl:
Finally got the IRATI stack to build on the Travis CI service. All is explained in the issue ticket](https://github.com/IRATI/stack/issues/1109).

This branch has the latest commits from the arcfire branch and gets a nice green tick mark beside it to show that the build passes.

The main point for review is that I had to remove the dependency on version 0.2.0 of librina in the configuration of [rinad](https://github.com/IRATI/stack/blob/1109-travis-ci/rinad/configure.ac#L117) and [rina-tools](https://github.com/IRATI/stack/blob/1109-travis-ci/rina-tools/configure.ac#L141). 

The system still checks that librina is there, but it does not check for a specific version. 

The reason, this check is not done, is due to the fact that librina uses [git-version-gen](https://github.com/IRATI/stack/blob/1109-travis-ci/librina/build-aux/git-version-gen) to create it's PACKAGE_VERSION, however in the Travis CI world the $VERSION that is generated by this system build is not passed as a global environment variable and thus the VERSION of librina gets set to UNKNOWN in it's package config file .pc. The knock on effect is that configure just bombs out when checking for librina at 0.2.0 or greater.
 
As stated with this change, the configure script still looks for `librina` but no version is specified.

Please review and let me know if any further changes are necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,13 @@ before_install:
   - sudo apt-get install git
   - sudo apt-get install g++
   - sudo apt-get install libssl-dev
-  #- sudo apt-get install protobuf-compiler 
-  #- sudo apt-get install libprotobuf-dev 
+  #- sudo apt-get install protobuf-compiler
+  #- sudo apt-get install libprotobuf-dev
   - sudo apt-get install hostapd
   - sudo apt-get install wpasupplicant
 
 before_script:
-  # Make a build and install directory 
+  # Make a build and install directory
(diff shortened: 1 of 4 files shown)


vmaffione:
@miguelpdl It's ok to remove dependency on that specific version of librina.
The current build system is needlessly complex and should be replaced with something way easier, with no separate packages.

PR #1180, opened 2017-12-13
rina-tools: iporinad: detect and react to closed sessions

vmaffione:
This patch also includes several changes that refactors iporinad
and the fdfwd lib.

### DIFF
--- a/rina-tools/src/rlite/Makefile.am
+++ b/rina-tools/src/rlite/Makefile.am
@@ -15,7 +15,7 @@ rina_echo_async_CPPFLAGS = $(LIBRINA_API_CFLAGS)
 
 rina_gw_SOURCES = rina-gw.cpp fdfwd.cpp
 rina_gw_LDADD = $(LIBRINA_API_LIBS)
-rina_gw_CPPFLAGS = $(LIBRINA_API_CFLAGS)
+rina_gw_CPPFLAGS = $(LIBRINA_API_CFLAGS) -std=c++11
 
 MOSTLYCLEANFILES =
 EXTRA_DIST =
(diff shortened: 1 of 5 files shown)


edugrasa:
Compiler fails due to a couple of things:

* the Makefile in rina-tools/src/rlite needs to specify c++11 when compiling the rina-gw
* when compiliing the iporinad file the compiler says it cannot find the <rina/cdap.hpp> header (including "cdap.hpp" as before works)

After compiling I've setup a simple scenario, killed one of the daemons but the other one does not react (flow is deallocated though). Maybe a problem in IRATI?

edugrasa:
I've investigated more.. the flow is deallocated, then the iporinad daemon tries to read the file descriptor associated to the flow, the io-device calls the KFA, who tries to recover the flow, but it doesn't exist and returns return -EBADF.

So the read syscall returns -EBADF, but I can see that iporinad issues another read syscall right away, which reports another error, and so on.. (until I kill the iporinad process). Maybe I should return another error?

vmaffione:
Sorry for the build issues, I just forgot to redo those changes, and I didn't realize the error because I only tried to compile iporinad on my system (that has also the headers installed).

Regarding the EBADF the way to go is that kernel returns 0 to indicate EOF. This is what rlite does, and it's the common way to indicate EOF.
As a side thing, it's not ok that fdfwd loops over an error, I beed to fix that.

edugrasa:
Ok, I'll fix that so it returns 0.

vmaffione:
Fixed the build issues, now it should build correctly.

edugrasa:
It still tries to read the file descriptor, even returning 0.. I'll check again.

edugrasa:
Ok, now it works, it was a problem with the poll implementation in the iodev, fixed!

PR #1181, opened 2017-12-14
read() syscall on iodev must return 0 (EOF) when the flow has been deallocated

edugrasa:


kernel/kfa.c
```diff
@@ -914,20 +914,20 @@ int kfa_flow_du_read(struct kfa  *instance,
 	if (!flow) {
 		LOG_ERR("There is no flow bound to port-id %d", id);
 		spin_unlock_bh(&instance->lock);
-		return -EBADF;
+		return 0;
```
vmaffione: Maybe we need this for now, but it's not technically correct: how do you distinguish the EOF condition from a random port-id provided by the user?

edugrasa: The user cannot specify a port-id directly when reading.. it issues a read syscall on a File descriptor. If the file descriptor does not exist, then it will never reach this code. If it does exist, the iodev instance has the mapping of the fd to the port-id, which is always correct. Then the iodev calls the KIPCM which calls the KFA.. So the case you mention can never happen, right?

vmaffione: (by user I meant a generic caller of the `kfa_flow_du_read` function)
However: yes, it makes sense with the assumption that "the function is called and there is not flow for this port-id" --> "flow was there and has been deallocated".
Just saying that if it's possible that the function is called with something that is not a port-id previously allocated then it's not correct.

vmaffione:
Do you see iporinad reacting now?

PR #1182, opened 2017-12-14
kernel: iodev: fixed poll

edugrasa:


vmaffione:
yes, indeed (https://github.com/vmaffione/rlite/blob/master/kernel/io-dev.c#L534-L539)

PR #1192, opened 2018-03-29
kernel: shim-eth-vlan: improve vertical flow control and remove qdiscs

vmaffione:
IRATI-specific qdisc has been removed so that now we are free to configure qdiscs.
Vertical flow control between the upper DIFs and the shim eth is implemented by means of a simple flow control window.


sandervrijders:
Great stuff guys!

dstaesse:
+1

edugrasa:
Nice work Vincenzo! I've tested it and seems to work ok. I'll merge it so that we can do more detailed experimentation in the testbeds.

PR #1193, opened 2018-04-10
Arcfire qta

edugrasa:
* Enhance QTA mux implementation: ease configuration, implement token-bucket policer, fix C/U mux, enhance documentation
* Fix RMT workflow (so that all PDUs go through the RMT policy)
* Fix deadlock in Flow allocator for local flows

### DIFF
--- a/common/serdes-utils.c
+++ b/common/serdes-utils.c
@@ -521,7 +521,8 @@ int flow_spec_serlen(const struct flow_spec * fspec)
 {
 	if (!fspec) return 0;
 
-	return 8 * sizeof(uint32_t) + sizeof(int32_t) + 3* sizeof(bool);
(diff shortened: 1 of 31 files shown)


vmaffione:
Sounds good, but any change to the rina api header must be discussed and done together across the prototypes. This is also valid for common applications. Otherwise we end up with code that is not interoperable.

vmaffione:
In this case no changes to the header, so that's ok. However, to avoid conflicts with the rinaperf upstream it would be better to do the change first in the upstream, and then replay those changes in irati. I can do the changes in the upstream and then align irati.

edugrasa:
Hi Vincenzo, ok, let's do it this way. Can I merge this PR and then we merge your changes on top?

vmaffione:
Yes please go ahead

PR #1196, opened 2018-04-19
rina-tools: upgrade iporinad, cdap and fdfwd

vmaffione:
This includes various fixes and support for -E/-L arguments for
iporinad (max loss and max delay).

### DIFF
--- a/rina-tools/src/rlite/iporinad.cpp
+++ b/rina-tools/src/rlite/iporinad.cpp
@@ -156,6 +156,10 @@ public:
     /* Enable verbose mode */
     int verbose = 0;
 
+    /* QoS parameters */
+    int max_delay = 0;
+    int max_loss  = RINA_FLOW_SPEC_LOSS_MAX;
+
     void start_workers();
     int setup();
     int main_loop();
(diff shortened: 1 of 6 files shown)


rina-tools/src/rlite/iporinad.cpp
```diff
@@ -1227,23 +1221,55 @@ IPoRINA::connect_to_remotes()
...
+}
+
 static void
 usage(void)
 {
     cout << "iporinad [OPTIONS]" << endl
          << "   -h : show this help" << endl
          << "   -c CONF_FILE: path to config file "
             "(default /etc/rina/iporinad.conf)"
-         << endl;
```
edugrasa: This endl should be there, otherwise the help option produces this output:

````
iporinad [OPTIONS]
   -h : show this help
   -c CONF_FILE: path to config file (default /etc/rina/iporinad.conf)   -L NUM : maximum loss probability introduced by the flow (NUM/10000)
   -E NUM : maximum delay introduced by the flow (microseconds)
   -v : be verbose
````

vmaffione: should be fixed now

edugrasa:
Works well now, tested with 2 nodes :)

PR #1198, opened 2018-04-23
Eraser apps: integrated traffic generation apps from ERASER project

edugrasa:
And added a couple of fixes:

* remove too detailed debug messages in librina
* fixed rina name separator at RINA API level


vmaffione:
what is ERASER?

edugrasa:
A project that was funded as part of January's FED4FIRE+ open call. It is a single partner project (UPC), and the goal is to carry out a RINA experiment in the FIRE testbeds using the tools developed/improved by ARCFIRE

vmaffione:
Sounds nice, thanks

PR #1204, opened 2018-05-23
Add protocol to www.irati.eu href to load correctly

brixen:
The anchor link href for www.irati.eu uses a relative syntax, causing the browser to try to load "http://irati.github.io/stack/www.irati.eu" instead. Adding the protocol so the href is "http://www.irati.eu" should fix the issue.


edugrasa:
Thanks for the PR!

PR #1322, opened 2020-05-08
exponentially increase cwnd during SLOW_START phase

kr1stj0n:
Shouldn't cwnd increased exponentially rather than linearly during slow_start phase?

### DIFF
--- a/plugins/dctcp/dtcp-ps-dctcp.c
+++ b/plugins/dctcp/dtcp-ps-dctcp.c
@@ -76,7 +76,7 @@ static int dctcp_rcvr_flow_control(struct dtcp_ps * ps, const struct pci * pci)
 		data->state = CONG_AVOID;
 	} else if (data->state == SLOW_START) {
 		/* Increase credit by one */
-		cwnd++;
+		cwnd = cwnd << 1;
 	} else {
 		/* CA state, increase by 1/cwnd */
 		data->dec_credit += DEC_PRECISION/cwnd;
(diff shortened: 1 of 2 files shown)


edugrasa:
Good catch, thanks Kristjon!

gravity64:
If those `rcvr_flow_control` callbacks are invoked once for each packet, I don't think the change is correct.
Incrementing the congestion window size by one unit (if it is measured in PDUs) or by one MSS (if it is measured in bytes) for each PDU that was correctly acked will already cause an exponential grow of the window. If you start for example with a window of 1 PDU, you will send a single PDU and wait for the ack, since the window is already full. When the ack arrives, you will increment the windows to 2, and then send 2 PDUs to make the window full again. Then you will receive 2 acks, and for each ack you will increment the window by 1 unit, which means that the new window will be 4 PDU large. Then you send 4 PDUs, and receive 4 acks, increasing the window size to 8. You see that the window grows exponentially, but still you are incrementing the window by 1 PDU for each received ack.
If you have ack coalescing you may adjust the computation by keeping into account the number of PDUs that the ACK control PDU actually acknowledged since the last ACK.

kr1stj0n:
I agree with you! This is what TCP Reno and others normally do, for instance.

However, as far as I checked the IRATI code, the congestion control policies are applied at the receiver side, and `rcvr_flow_control` is invoked whenever the receiver gets a PDU. It increases the `rcvr_credit`, and thus the `right_window_edge`, which is advertised at the sender. The sender can send only when the receiver window is not closed. @edugrasa, please correct me if I missed something.

edugrasa:
It is implemented just as you explained Kristjon

PR #1344, opened 2021-05-19
5.10 fixes

fdgonthier:
This fixes compilation of the IRATI kernel modules for Linux 5.10 kernel

This is just an RFC PR as this is **untested**. The patch is enough for the modules to compile and load but I I do not know how to configure my sample RINA setup with encryption so that this code path get used. Suggestion are welcome.

### DIFF
--- a/kernel/dtcp.c
+++ b/kernel/dtcp.c
@@ -23,6 +23,7 @@
 #define RINA_PREFIX "dtcp"
 
 #include <linux/delay.h>
+#include <linux/ktime.h>
 
 #include "logs.h"
 #include "utils.h"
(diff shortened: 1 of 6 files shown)


TRIA:
The CI failures are on kernels 4.1 and 4.2, which had not yet incorporated the new crypto API.  Those kernels are ancient, and could be removed from the CI with little loss.

edugrasa:
Hi, 

Thanks for the fix proposal! However it does not comply with the practice we have been following to support multiple kernel version (which is not written anywhere, I know and sorry for that :) ).

The trick is to use #if, #else, #endif primitives that take into account the kernel version, like in https://github.com/IRATI/stack/blob/master/kernel/ipcps/shim-tcp-udp.c#L724

Could you update your pull request so that the new code you added gets used for versions 4.3 up and the old code for versions 4.2 down? Thanks a lot!

fdgonthier:
Alright, I will update the code, no problem.

But I would also like to know how to test this feature. I couldn't find any configuration example about this.

edugrasa:
Hi,

You can use the tutorial in https://github.com/IRATI/stack/wiki/Tutorial-3:-Security-provider-net to test this feature. You don't need to deploy the whole scenario depicted in the tutorial, it will be enough with the red DIF (access. DIF) and the "Customer border Router" and "Provider border router 1" systems (so a couple of VMs are enough for testing this).

Cheers,

Eduard

edugrasa:
Closing this PR. The code to support 5.5+ kernels has been already merged in; in case something does not work please let me now :)

PR #1351, opened 2021-07-30
protobuf: change ByteSize to ByteSizeLong

rockrush:
According to github.com/protocolbuffers/protobuf:CHANGES.txt, ByteSize() and SpaceUsed()
are deprecated since protobuf 3.4.0 at 2017-08-14, protobuf suggested to use ByteSizeLong()
and SpaceUsedLong() instead.

With protobuf 3.12.4 and GCC 10.2.1, this already blocked 'stack' from been built.


rockrush:
Seems to be out of interest

rockrush:
Seems to be out of interest

PR #1361, opened 2021-09-13
Missing spin_unlock before return.

fdgonthier:
This statement was obviously missing and I think this is what caused the last lock ups that I've seen. Before going trigger-happy about closing the associated bug, though, I think I'll use the module for a bit longer.


edugrasa:
@fdgonthier should have seen this before merging the pull request, I reviewed the code too fast. Sorry for that!

PR #1369, opened 2021-12-20
ARP request timeout

fdgonthier:
The timeout is somewhat arbitrarily set to 5000 milliseconds but can be changed using the shim
    configuration file, or using DebugFS. A failure to receive an ARP before the timer expires will lead
    the flow creation to fail and the ressources allocated to it (ie: the port number) to be freed.
    
    Also added are a few features available using DebugFS. In the 'arp826' directory, any data written
    to 'force-timeout' will lead all pending ARP requests to timeout. The 'resolutions' file will show
    data about all the pending ARP requests.


fdgonthier:
Reworking my pull requests.

PR #1372, opened 2022-01-12
Kernel bugfix + another feature

fdgonthier:
This PR includes one important bugfix + another feature that is in another PR.


fdgonthier:
I'm in the process of making another set of pull requests, from a cleaner slate. The PRs will include this set of patch so I'll close this one for now.

